PR #99, opened 2014-06-26
DATAJPA-562 - Support for pluggable generic post-processing of JPA Query results.

thomasdarimont:
We introduced the JpaResultPostProcessor interface to provide the  infrastructure to be able to post-process the results of entityManager.getResultList() or entityManager.getSingleResult() invocations.
This enables us to manipulate the results like automatically detaching them from the PersistenceContext or masking values etc via pluggable strategies.
By default a NoopJpaResultPostProcessor is used that returns the results as is. We also provide a DetachingJpaResultPostProcessor that detaches all returned domain-objects from the associated persistence-context.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/support/DetachingJpaResultPostProcessor.java
+++ b/src/main/java/org/springframework/data/jpa/repository/support/DetachingJpaResultPostProcessor.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2008-2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


RobzInadE:
Why was this closed? This is such an invaluable thing for me to have in my project. It's freaking impossible to have a clean usable interceptor to a repository without having to use AOP which just causes headache and unnecessary typing.

thomasdarimont:
Hi @RobzInadE,

here is some additional context: https://jira.spring.io/browse/DATAJPA-562
I also think that this would still be a very useful feature.

Cheers,
Thomas

RobzInadE:
I see. I tested your solution, however calls to non-implemented "magical" repository methods were not caught in any of the SimpleJpaRepository class. I kind of need something like this in my project like I said. Have you got any other ideas of implementing a similar solution (that also applies to repository methods) ? 

Regards

PR #110, opened 2014-10-23
DATAJPA-617 - Avoid potential transaction roll backs during named query lookups.

thomasdarimont:
We now use a dedicated EntityManager for the named query lookup in NamedQuery.hasNamedQuery(..). Previously we used the given EntityManager to lookup the named query which cause transaction to be rolled back due to the Exception thrown in the case that the named query couldn't be found. 
Added fast path to NamedQuery.lookupFrom(..) that now uses NamedQuery.hasNamedQuery(...) to test if a query exists or not.
Pulled the check for the presence of a namedCountQuery from doCreateCountQuery up to the NamedCountQuery constructor. This makes sure that the check is only performed once.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/NamedQuery.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/NamedQuery.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2008-2013 the original author or authors.
+ * Copyright 2008-2014 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


mirackle-spb:
Looks like problem is still there when using spring-data-jpa with JTA transaction manager like bitronix or atomikos. Transaction is marked as rollbackonly after lookup is failed.

mirackle-spb:
Looks line obtain dedicated EM is not enough and need to create specific transaction for lookup method.

schauder:
@mirackle-spb if you see the same or a similar issue as this one please create a new issue because this one is closed.

PR #122, opened 2014-11-14
DATAJPA-629 - Allow SpEL template expressions in combination with parameter expressions in @Query.

thomasdarimont:
Previously SpEL template expressions like #{#entityName} could not be used in conjunction with parameter expressions in @Query because the SpEL template parser tried to evaluate the dynamic parameter expressions as well. We now mask the parameters prior to evaluating the SpEL template expression.


src/main/java/org/springframework/data/jpa/repository/query/ExpressionBasedStringQuery.java
```diff
@@ -35,6 +37,13 @@
  */
 class ExpressionBasedStringQuery extends StringQuery {
 
+	private static final String EXPRESSION_PARAMETER = "?#{";
```
mgurevin: this works only `:?{param}`
but there are another syntax: `:#{param}`

PR #129, opened 2014-12-22
DATAJPA-412 - Add refresh method to SimpleJPARepository.

emckissick:
Exposed entity manager refresh method the JpaRepository interface and an implementation to SimpleJpaRepository.  Also added a locked mode type check for refresh. 

Cleaned up commits from original branch.


olivergierke:
The original ticket has been resolved as won't fix. Closing this one.

PR #130, opened 2015-01-16
DATAJPA-652 - Added support for REF_CURSOR output parameters for procedures.

thomasdarimont:
We now support detecting output parameters for stored-procedures with ParameterMode.REF_CURSOR. Previously we only considered parameters with OUT or INOUT mode as output parameters.

Note that since it is a bit tricky to setup a proper test case with hsql / eclipselink, I had to verify this with a standalone spring boot app:
https://gist.github.com/thomasdarimont/129bc15d0ccc459610c2
A proper test case will follow.


mkolesnikov:
This is still an issue. REF_CURSOR should be retrieved with `getResultList()` but Spring-Data-JPA tries to get it with 'getOutputParamenterValue()' and it fails. `getResultList()` works fine with Hibernate 5.1.0. and Spring-Data-JPA 1.9.4

vishnudevk:
I am experiencing this issue with Hibernate 5.2.10.Final

PR #132, opened 2015-01-23
DATAJPA-666 - Limit maxResults in SingleEntityExecution to 1.

thomasdarimont:
We now set maxResults to 1 on a JPA Query in case of a SingleEntityExecution.
Setting maxResults to 1 has the effect that custom queries via @Query that return multiple results which previously resulted in a TooManyResultsException will now 
return the first result.

This enables some advanced use cases such as:

``` java
@Query("FROM Image ORDER BY rand()")
Image findFirstRandomImage();
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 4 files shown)


schauder:
We won't merge this since it would break compatibility. See Jira for more details.

PR #136, opened 2015-02-24
DATAJPA-677 - Add support for Java 8 Stream in Repository finder methods.

thomasdarimont:
Just a quick PoC for Java 8 Stream support for JPA Repository finder methods based on hibernate.
I think we could add (provider specific) support for Java 8 Streams here but we probably have to fiddle a bit with the particular PersistenceProvider internals to make it work.

This is builds on top of spring-data-commons DATACMNS-650-SNAPSHOT.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
@@ -15,8 +15,13 @@
  */
 package org.springframework.data.jpa.repository.query;
 
+import java.lang.reflect.Field;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.StreamSupport;
 
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
```diff
@@ -298,4 +309,48 @@ protected Object doExecute(AbstractJpaQuery jpaQuery, Object[] values) {
 			return storedProcedureJpaQuery.extractOutputValue(storedProcedure);
 		}
 	}
+
+	static class StreamExecution extends JpaQueryExecution {
+
+		@Override
+		protected Object doExecute(AbstractJpaQuery query, Object[] values) {
+
+			Query jpaQuery = query.createQuery(values);
```
thomasdarimont: This is just a raw sketch .. for a real implementation we need to abstract the Provider details ...

src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
```diff
@@ -298,4 +309,48 @@ protected Object doExecute(AbstractJpaQuery jpaQuery, Object[] values) {
...
+
+			QueryImpl<Object> queryImpl = (QueryImpl<Object>) jpaQuery;
+
+			Field field = ReflectionUtils.findField(QueryImpl.class, "query");
+			ReflectionUtils.makeAccessible(field);
+
+			org.hibernate.Query qry = (org.hibernate.Query) ReflectionUtils.getField(field, queryImpl);
+
+			final ScrollableResults scrollableResults = qry.scroll(ScrollMode.FORWARD_ONLY);
```
thomasdarimont: We need to find a better way to construct `ScrollableResults`

olivergierke:
Already merged and released.

PR #137, opened 2015-03-09
DATAJPA-689 - Allow @EntityGraph on findOne method of CrudRepository.

thomasdarimont:
We now honor @EntityGraph definitions on findOne methods - previously we only supported it on methods that created a Query explicitly.
Extracted tryGetFetchGraphHints method from tryConfigureFetchGraph method in Jpa21Utils to allow EntityGraph hints to be used in SimpleJpaRepository#findOne.
Construction of query hints from context information in SimpleJpaRepository is now performed via the getQueryHints method.
Adjusted QueryDslJpaRepository to use getQueryHints as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>1.8.0.BUILD-SNAPSHOT</version>
+	<version>1.8.0.DATAJPA-689-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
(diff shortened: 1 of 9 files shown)


olivergierke:
It feels like we run into a plethora of static helper methods that all seem similar but do slightly different things. If the entity graph lookup can be combined with the query hints, can't we make sure we actually need a single public static helper method that does all of the necessary stuff to a query? 

It feels like we could easily get lost in partially applying some stuff to `Query` objects and forgetting some other stuff. We should make sure that the CRUD methods use the same API as the query execution does, which doesn't seem to be the case currently.

PR #140, opened 2015-04-01
DATAJPA-696 - Support locally declared EntityGraphs for repository finder methods.

thomasdarimont:
We now support the specification of dynamic EntityGraphs on repository finder methods by allowing to specify the fetch graph paths via the “attributePaths” attribute on the EntityGraph annotation.
Previously users had to specify the EntityGraph specifications upfront via `@NamedEntityGraph` on the entity-class.

Configured eclipse link tests to use dynamic weaving, required for dynamic entity fetch graphs.
Upgraded eclipse link to version 2.5.2 for full fetch-graph support.
Fixed typo in CrudMethodMetadataPopulatingMethodInterceptor.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
+++ b/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
@@ -15,8 +15,17 @@
  */
 package org.springframework.data.jpa.provider;
 
-import static org.springframework.data.jpa.provider.JpaClassUtils.*;
-import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.*;
+import static org.springframework.data.jpa.provider.JpaClassUtils.isEntityManagerOfType;
+import static org.springframework.data.jpa.provider.JpaClassUtils.isMetamodelOfType;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.ECLIPSELINK_ENTITY_MANAGER_INTERFACE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.ECLIPSELINK_JPA_METAMODEL_TYPE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.GENERIC_JPA_ENTITY_MANAGER_INTERFACE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.HIBERNATE43_ENTITY_MANAGER_INTERFACE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.HIBERNATE43_JPA_METAMODEL_TYPE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.HIBERNATE_ENTITY_MANAGER_INTERFACE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.HIBERNATE_JPA_METAMODEL_TYPE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.OPENJPA_ENTITY_MANAGER_INTERFACE;
+import static org.springframework.data.jpa.provider.PersistenceProvider.Constants.OPENJPA_JPA_METAMODEL_TYPE;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -24,8 +33,10 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
+import javax.persistence.EntityGraph;
 import javax.persistence.EntityManager;
 import javax.persistence.Query;
+import javax.persistence.Subgraph;
 import javax.persistence.metamodel.Metamodel;
 
 import org.apache.openjpa.enhance.PersistenceCapable;
@@ -42,10 +53,12 @@ import org.hibernate.ScrollableResults;
 import org.hibernate.ejb.HibernateQuery;
 import org.hibernate.proxy.HibernateProxy;
 import org.springframework.beans.DirectFieldAccessor;
+import org.springframework.data.jpa.repository.query.JpaEntityGraph;
 import org.springframework.data.util.CloseableIterator;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * Enumeration representing persistence providers to be used.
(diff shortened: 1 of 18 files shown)


src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
```diff
@@ -15,8 +15,17 @@
  */
 package org.springframework.data.jpa.provider;
 
-import static org.springframework.data.jpa.provider.JpaClassUtils.*;
```
olivergierke: Please revert to star imports.

src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
```diff
@@ -355,6 +368,69 @@ public String getCountQueryPlaceholder() {
 	}
 
 	/**
+	 * Creates a dynamic {@link EntityGraph} from the given {@link JpaEntityGraph} information.
```
olivergierke: What's the reason these methods live in `PersistenceProvider`? There's nothing provider specific in here. Maybe a left over from the state of using provider specific API?

src/main/java/org/springframework/data/jpa/repository/query/Jpa21Utils.java
```diff
@@ -83,17 +86,25 @@ private Jpa21Utils() {
...
 
 		Assert.isTrue(JPA21_AVAILABLE, "The EntityGraph-Feature requires at least a JPA 2.1 persistence provider!");
 		Assert.isTrue(GET_ENTITY_GRAPH_METHOD != null,
 				"It seems that you have the JPA 2.1 API but a JPA 2.0 implementation on the classpath!");
 
-		return em.getEntityGraph(jpaEntityGraph.getName());
+		try {
+			// first check whether an entityGraph with that name is already registered.
+			return em.getEntityGraph(jpaEntityGraph.getName());
+		} catch (Exception ex) {
```
olivergierke: There's a few issues I suspect popping up here: the exception being thrown will render the `EntityManager` unusable as defined by the JPA spec (see last paragraph on page 79, JPA 2.1 spec). Thus we either need to create a separate `EntityManager` instance for the check or rather avoid the lookup altogether.

It generally feels a bit wrong to basically register something within in the `EntityManager` in a method that seems to only lookup stuff. Can't we rather change the general model to do the registration once at bootstrap time and remember the name of the entity graph. On an actual method execution we then can look up the graph by name if present or skip this step entirely.

src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
```diff
@@ -355,6 +368,69 @@ public String getCountQueryPlaceholder() {
...
+	/**
+	 * Configures the given {@link EntityGraph} with the fetch graph information stored in {@link JpaEntityGraph}.
+	 * 
+	 * @param jpaEntityGraph
+	 * @param entityGraph
+	 */
+	/* visible for testing */
+	void configureFetchGraphFrom(JpaEntityGraph jpaEntityGraph, EntityGraph<?> entityGraph) {
+
+		String[] attributePaths = jpaEntityGraph.getAttributePaths().clone();
```
olivergierke: Should we rather use a `List` here?

src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
```diff
@@ -355,6 +368,69 @@ public String getCountQueryPlaceholder() {
...
+
+		// sort to ensure that the intermediate entity subgraphs are created accordingly.
+		Arrays.sort(attributePaths);
+
+		// we build the entity graph based on the paths with highest depth first
+		for (int i = attributePaths.length - 1; i >= 0; i--) {
```
olivergierke: Does it make sense to use `PropertyPath` or `PersistentPropertyPath` here to get some kind of verification whether the path segments actually refer to existing properties?

src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadata.java
```diff
@@ -45,10 +47,18 @@
...
-	JpaEntityGraph getEntityGraph();
+	EntityGraph getEntityGraph();
+	
+	/**
+	 * Returns the {@link Method} to be used.
+	 * 
+	 * @return
+	 * @since 1.9
+	 */
+	Method getMethod();
```
olivergierke: This one feels weird, as the sole purpose of the `CrudMethodMetdata` is for clients not having to deal with the method being invoked. See more comments on that on the spot in `SimpleJpaRepository` where this is used.

src/main/java/org/springframework/data/jpa/repository/query/Jpa21Utils.java
thomasdarimont: we can also work around that... 
via:

```java
if(!jpaEntityGraph.isDynamic()){
  return em.getEntityGraph(...)
}

return ...createDynamicEntityGraph(...)
```

src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadataPostProcessor.java
```diff
@@ -85,7 +86,7 @@ public CrudMethodMetadata getLockMetadataProvider() {
 	 * @author Oliver Gierke
 	 * @author Thomas Darimont
 	 */
-	static enum CrudMethodMetadataPopulatingMethodIntercceptor implements MethodInterceptor {
+	static enum CrudMethodMetadataPopulatingMethodInterceptor implements MethodInterceptor {
```
olivergierke: The `RepositoryProxyPostProcessor` creating this `MethodInterceptor` already knows about a `RepositoryInformation`. If we turned the enum into a class, we could actually wire the `EntityInformation` contained in the `RepositoryInformation` to the instance about to be created and thus determine the default entity graph name internally without having the `SimpleJpaRepository` manually assemble it by obtaining information from the `CrudMethodMetadata` and then eventually forwarding to the `JpaEntityGraph`.

olivergierke:
So generally speaking I think we should improve the way the locally declared graphs are registered. Currently there seems to be some kind of "register if not already registered, but if so - use it" approach with doesn't play nicely with JPA's rather strict exception throwing/handling policy.

I'd prefer an approach that registers the graph at bootstrap time and then basically only returns the graph name to the infrastructure so that to the client code trying to apply the graph, it's basically about looking up the graph or not.

WDYT?

src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadataPostProcessor.java
thomasdarimont: Yep - thought about that too - ok I give it a shot.

thomasdarimont:
> I'd prefer an approach that registers the graph at bootstrap time and then basically only returns the
> graph name to the infrastructure so that to the client code trying to apply the graph, it's basically about
> looking up the graph or not.

I don't know whether this is possible since dynamic entityGraphs don't seem to support a name :-/

PR #142, opened 2015-04-10
DATAJPA-704 Allow specifying/providing query strings programatically

nickguletskii:
https://jira.spring.io/browse/DATAJPA-704
## Summary

Query providers are classes that implement the JpaQueryProvider interface. They serve as a way to set the queries at runtime, meaning that you can load queries from files, build them using a DSL, etc...
instead of putting them into the Query annotations as constants. This is better than using named queries because named queries force you to use either the NamedQuery annotation (same problem as above) or orm.xml (editing SQL in XML is inconvenient).
## Notes

At first, I wanted to implement query loading using Spring resource locations, but I realised that the current architecture of the internals of Spring Data is going to make that quite difficult to do, since this would require implementing ResourceLoaderAware.

I also considered just adding support for custom query annotations, but 
## Examples

Simple query provider that also provides the count query and projection:

``` java
static class ValidCountQueryProvider implements JpaQueryProvider, JpaCountQueryProvider {

    @Override
    public String getQuery() {
        return "query";
    }

    @Override
    public String getCountQuery() {
        return "countQuery";
    }

    @Override
    public String getCountProjection() {
        return "countProjection";
    }
}

@Query(queryProvider = ValidCountQueryProvider.class)
List<User> queryMethodUsingCountQueryProvider(String lastname);
```

The MethodAwareQueryProvider interface can be implemented to make a more powerful, generic query provider:

``` java
static class MethodAwareQueryProvider implements JpaQueryProvider, MethodAwareJpaQueryProvider {

    private Method method;

    // Will be called before getQuery and other methods
    @Override
    public void setMethod(Method method) {
            this.method = method;
    }

    @Override
    public String getQuery() {
           ...
    }

}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>1.9.0.BUILD-SNAPSHOT</version>
+	<version>1.9.0.DATAJPA-704-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
 	<description>Spring Data module for JPA repositories.</description>
--- a/src/main/java/org/springframework/data/jpa/repository/JpaCountQueryProvider.java
+++ b/src/main/java/org/springframework/data/jpa/repository/JpaCountQueryProvider.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2008-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


pivotal-issuemaster:
@nickguletskii Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=142) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

PR #149, opened 2015-06-09
Fix typo in readme

johnktims:


olivergierke:
Attention to detail, I like! :) Thanks for that one!

PR #150, opened 2015-07-01
DATAJPA-218 - Support extracting parameters from a bean parameter.

thomasdarimont:
Added prototypic support for query by example queries to 
SimpleJpaRepositories. Clients can use an Example Object to
wrap an existing prototype entity instance that will be used to 
derive a query from.
The example wrapper allows us to customize the parameter extraction / generation 
from the provided prototype bean.

Would be great if we could unify this effort with the infrastructure
from DATAMONGO-1245.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>1.9.0.BUILD-SNAPSHOT</version>
+	<version>1.9.0.DATAJPA-218-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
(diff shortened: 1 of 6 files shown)


thomasdarimont:
Potential things to improve / think about are:
- Provide a common shared Example wrapper in Spring Data Commons.
- Support nested attribute paths for exclusion.
- Consider an alternative style to express example predicates:
```java
 userRepository.findByExample(new Example(){
    String name = "thomas";
    String countryCode = "de";
    Boolean tired = true;
});
```
- Provide query by example at ``CrudRepository`` level.

ripper2hl:
Why no accept this pull request?, better approach ? , the querys by Example its a great feature :+1:

schauder:
This is superseeded by #164

PR #152, opened 2015-07-31
DATAJPA-382 - Demonstrate that dynamic SpEL parameters work in update queries.

thomasdarimont:
Added test case to AbstractAuditingViaJavaConfigRepositoryTests that uses the SampleEvaluationContextExtension
to demonstarte access of "security" context objects as well as dynamic SpEL Expression binding.

Needed to add FixedDate helper class to ease checks for modification time stamps.


olivergierke:
That's merged!

PR #154, opened 2015-08-13
1.7.x  An array type of return value does not return the desired results

jaiiye:
spring-boot-1.25(spring-data-jpa-1.72+hibernate-4.3.10)
An array type of return value does not return the desired results,for example: 
@Query("SELECT COUNT(cd.id),SUM(cd.jyje),SUM(cd.sxf),SUM(cd.jsje) FROM ConsumerDetails cd WHERE cd.shh IN :merchantCodes AND cd.jsdate >= :startDate AND cd.jsdate < :endDate")
    Object[] countByCriteria(@Param("merchantCodes") List<String> merchantCodes,
            @Param("startDate") String startDate,
            @Param("endDate") String endDate);
Returns the Object[][].

entity:
private Integer id;
private Double jyje;
private Double sxf;
private Double jsje;
...


olivergierke:
Would you mind cleaning up this one a bit? It contains quite a few unrelated commits.

PR #159, opened 2015-12-03
Replacing bean name "foo" with more meaningful value

Nugs:
Happened across this one by accident. Just a general piece of housekeeping - a bean probably should have a meaningful name!


olivergierke:
That's merged, thanks!

PR #162, opened 2016-01-19
DATAJPA-849 - Fix test case in README

is-alex:


olivergierke:
That's merged, thanks!

PR #163, opened 2016-01-20
Add documentation on how to use the @EntityListeners annotation

wimdeblauwe:
I was trying out the auditing support, but I did not get it working from the documentation alone. Luckily, I found http://www.petrikainulainen.net/programming/spring-framework/spring-data-jpa-tutorial-auditing-part-one/ that explained I neede the `@EntityListeners` annotation (I am not using an `orm.xml`). I think it would be good to add this to the official docs as well.

### DIFF
--- a/src/main/asciidoc/jpa.adoc
+++ b/src/main/asciidoc/jpa.adoc
@@ -744,6 +744,19 @@ Note that the auditing feature requires `spring-aspects.jar` to be on the classp
 ----
 ====
 
+You can also enable the `AuditingEntityListener` per entity using the `@EntityListeners` annotation:
+
+====
+[source, java]
+----
+@Entity
+@EntityListeners(AuditingEntityListener.class)
+public class MyEntity {
+
+}
+----
+====
+
 Now activating auditing functionality is just a matter of adding the Spring Data JPA `auditing` namespace element to your configuration:
 
 .Activating auditing using XML configuration


olivergierke:
That's merged, thanks!

halower:
I have an audit interceptor very similar to this : http://www.mkyong.com/hibernate/hibernate-interceptor-example-audit-log/

Whenever i do in my service method:
```

 User user = this.userRepository.findOne(id);
 user.setX(x);
 user.setY(y);
I get an audit just fine! works perfect.
```

But when i use a query from whithin the respository like so:

```
UserRepository {

   @Query("UPDATE User u SET u.password = ?2 WHERE u.id = ?1 ")
   @Modifying
   public void updatePassword(Long userId, String encodedPassword);

}  
```
I always thought that since spring-boot-jpa supports @Query, it definitely supports auditing. However, the reality is very bad. The official documentation is not clear enough, so I have dug a big pit.，because my project is nearing the end, only to find this problem, so it is impossible to switch other ways, do not know how to solve it, or alternatives，I hope everyone can help me, I will wait for results online.

pivotal-issuemaster:
@wimdeblauwe Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=163)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=163) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@wimdeblauwe Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=163)!

wimdeblauwe:
@halower I suggest to ask this kind of question on StackOverflow as your comment is not directly related to this PR.

PR #165, opened 2016-02-15
DATAJPA-862 - fixed Query parameter index in adoc.

outofrange:
One Query annotation used an parameter index of ?0 instead of ?1.


olivergierke:
That's merged, thanks!

PR #171, opened 2016-04-11
DATAJPA-888 - Clarify usage of native queries for pagination

michael-simons:
This commit changes the note of using native queries for pagination and adds an example on how to manually define a count query. Also clarify that pagination using native queries is supported, dynamic sorting not.

This adresses questions like this one http://stackoverflow.com/questions/28529584/spring-data-why-its-not-possible-to-have-paging-with-native-query/36520935#36520935.

### DIFF
--- a/src/main/asciidoc/jpa.adoc
+++ b/src/main/asciidoc/jpa.adoc
@@ -249,7 +249,9 @@ public interface UserRepository extends JpaRepository<User, Long> {
 ----
 ====
 
-Using advanced `LIKE` expressionsThe query execution mechanism for manually defined queries using @Query allow the definition of advanced `LIKE` expressions inside the query definition.
+==== Using advanced `LIKE` expressions
+
+The query execution mechanism for manually defined queries using @Query allows the definition of advanced `LIKE` expressions inside the query definition.
 
 .Advanced like-expressions in @Query
 ====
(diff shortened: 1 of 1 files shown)


olivergierke:
That's merged and backported. Thanks!

pramoth:
Is Spring Data JPA generate SQL for query paging? I ask because I don't know how to specify MYSQL LIMIT parameter or Oracle ROWNUM. I think Spring data not generate code right?

michael-simons:
Hi @pramoth I don't understand your question. If you let your repository extend `PagingAndSortingRepository` or define an interface method like `Page<Stuff> findAllBySomething(String something, Pageable pageable)` then paginated queries are used. You pass your limit through the `size`field of a [PageRequest](http://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/PageRequest.html).

Spring Data either generates the correct JPQL query then for you which is in turn transformed by your JPA Provider (i.e. Hibernate, OpenJPA or EclipseLink) to the correct SQL code. On the other hand you can provide a custom named query or even a custom named native query. My commit in the docs is about the last point. Paginated queries are supported for native queries in partial, including offsetting and limiting but not sorting.

pramoth:
Hi @michael-simons Sorry for my english. 
In my question above I mean "Who is responsible to generate offsetting sql?".
After you answer me,I realize that it is responsibility of JPA provider to set offsetting sql.
Great job. Help me a lot.
Thank you very much.

PR #173, opened 2016-05-31
Add note about named parameters and @Param

michael-simons:
Clarifies that the use of @Param is optional when on Java 8 and compiling with -parameters flag.

Just a small addition, but hopefully useful in the docs and for those who want to keep 3rd party annotations to a minimum in repositories.


olivergierke:
That's merged, thanks!

PR #174, opened 2016-06-29
DATAJPA-912 - Optimize paged query execution.

mp911de:
We execute paged queries now in an optimized way. The data is obtained for each paged execution but the count query is deferred. We determine the total from the pageable and the results in which we don't hit the page size bounds.Such cases are if results are less than a full page without offset or results are greater `0` and less than a full page with offset. 

In all other cases we issue an additional count query.

----

Ticket: [DATAJPA-912](https://jira.spring.io/browse/DATAJPA-912)
Related tickets: [DATAJPA-124](https://jira.spring.io/browse/DATAJPA-124), [DATAJPA-477](https://jira.spring.io/browse/DATAJPA-477), [DATACMNS-884](https://jira.spring.io/browse/DATACMNS-884)
Requires: https://github.com/spring-projects/spring-data-commons/pull/169

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>1.11.0.BUILD-SNAPSHOT</version>
+	<version>1.11.0.DATAJPA-912-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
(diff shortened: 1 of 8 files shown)


olivergierke:
Shall we move `PageableExecutionSupport` to Spring Data Commons?

mp911de:
That would help other modules to benefit from the paging.

mp911de:
PR updated, adopted `PageableExecutionUtils` from Spring Data Commons.

PR #175, opened 2016-07-04
DATAJPA-923 - Query by Example using OR operator - disjunction (instead of AND)

sergiowww:
getPredicateDisjunction, which joins all restrictions using operator OR
instead of AND. The method getPredicate remains with the same behavior
as before the refactoring (using AND operator).


olivergierke:
Thanks for the proposal but I think we're going to go a slightly different route here.

PR #177, opened 2016-08-02
JpaClassUtils.isOfType performance improvement

mmimica:
Throwing exception is slow.


olivergierke:
Care to open a ticket and be more explicit about why you think this change is necessary? Did you run into performance problems? Do you have benchmarks that the change causes a significant improvement?

olivergierke:
Nevermind, looks like we've already done that and applied a fix in [DATAJPA-941](https://jira.spring.io/browse/DATAJPA-941).

PR #178, opened 2016-08-09
DATAJPA-948 - Provide extensibility of query creation

lordlothar99:
changed some visibilities to public


jeacott1:
this has been sitting here a long time. any reason its not merged?

andreasgebauer:
Would also like to see this merged but there is a conflict now

schauder:
We won't implement this as it currently is proposed. 
But I created a separate issue in an attempt to come up with a maintainable API for this kind of stuff: DATAJPA-1551

PR #179, opened 2016-08-19
DATAJPA-953 - Fix typo in reference documentation

icoderman:
Added escape char for avoiding replacement <= with &#8656; by asciidoctor library.


olivergierke:
That's merged, thanks!

PR #182, opened 2016-09-28
DATAJPA-790 - QueryException when applying @EntityGraph on findAll with QueryDSL

jntakpe:
This PR just create count query in a separate method without applying query hints


olivergierke:
That's polished, merged and back-ported.

PR #187, opened 2017-01-13
Fix DATAJPA-1044

marclefrancois:
Use a distinct count Query if the passed in query is distinct


olivergierke:
That's been polished and merged.

marclefrancois:
@olivergierke Big thanks for the quick turn around

PR #188, opened 2017-01-20
DATAJPA-1041 - Append ad hoc EntityGraph subgraphs properly. 

christophstrobl:
We now make sure to append instead of recreate `SubGraphs` correctly when `AttributeNodes` already exist for a given property or dot path. This change fixes errors when creating ad hoc fetch graphs like below where multiple properties are loaded via a `SubGraph` on the same type.

```java
@EntityGraph(attributePaths = { "colleagues.roles", "colleagues.colleagues" })
```

Additionally we fixed issues with the creation of "deep" ad hoc fetch graphs via `@EntityGraph` so that

```java
@EntityGraph(attributePaths = { "roles", "colleagues.roles", "colleagues.colleagues.roles" })
```

can be used as a short form of

```java
@NamedEntityGraph(name = "User.deepGraph",
        attributeNodes = {
                @NamedAttributeNode("roles"),
                @NamedAttributeNode(value="colleagues", subgraph = "User.colleagues")
        },
        subgraphs = {
                @NamedSubgraph(name = "User.colleagues", attributeNodes = {
                        @NamedAttributeNode("roles"),
                        @NamedAttributeNode(value = "colleagues", subgraph = "User.colleaguesOfColleagues")
                }),
                @NamedSubgraph(name="User.colleaguesOfColleagues", attributeNodes = {
                        @NamedAttributeNode("roles"),
                })

        })
})
```

However we had to disable EclipseLink tests for this one since there seems to be a glitch when calling `EntityManager.clear()` prior to applying the _fetch-/loadgraph_ causing the properties not to be in the correct load state. Omitting `EntityManager.clear()` is not an option since the load state is always `LOADED` even when leaving out query hints.

---

Supersedes: #186

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>1.11.0.BUILD-SNAPSHOT</version>
+	<version>1.11.0.DATAJPA-1041-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
 	<description>Spring Data module for JPA repositories.</description>
--- a/src/main/java/org/springframework/data/jpa/repository/query/Jpa21Utils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/Jpa21Utils.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import javax.persistence.AttributeNode;
 import javax.persistence.EntityGraph;
 import javax.persistence.EntityManager;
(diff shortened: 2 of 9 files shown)


bkempe:
This PR was closed, the [JIRA issue marked as fixed in Version 1.11 GA](https://jira.spring.io/browse/DATAJPA-1041) but the [commit that made it into that release](https://github.com/spring-projects/spring-data-jpa/commit/db4d0a0fa0bac51036fcc7dbb01f718240ff4366) does not appear to fix the issue (as opposed to this PR).

olivergierke:
Please file a new issue with details what you think is not fixed (best through a tiny sample that shows what's not working).

PR #194, opened 2017-03-29
DATAJPA-1080 - Use From instead of Root to sort Join

speralta:
Use From instead of Root in QueryUtils.

This PR for master


schauder:
That's merged. Thanks.

PR #197, opened 2017-04-22
DATAJPA-1098 - javadoc update.

rzymek:
Change javadoc of `Specification.toPredicate()`.
The javadoc was misleading. The method may return null as described in https://jira.spring.io/browse/DATAJPA-300


schauder:
I created and Jira issue for this so it appears in release notes and everything. Added the commit to all the relevant branches.

Thanks.

PR #198, opened 2017-04-24
DATAJPA-1099 Fix Travis builds

schauder:
Travis builds seemed to be failing due to a broken reference to codehaus.org but the actual reason was a missing snapshot dependency.

Fixed both issues by deleting Travis' settings.xml and upgrading the snapshot dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
 
 		<dist.key>DATAJPA</dist.key>
 
-		<eclipselink>2.6.2</eclipselink>
+		<eclipselink>2.6.4</eclipselink>
 		<hibernate>5.2.9.Final</hibernate>
 		<hsqldb1>1.8.0.10</hsqldb1>
 		<jpa>2.0.0</jpa>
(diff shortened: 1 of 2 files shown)


olivergierke:
I'd love to see the dependency upgrade in a separate ticket. Probably fixed before fixing Travis.

pom.xml
```diff
@@ -48,7 +48,7 @@
 		<profile>
 			<id>eclipselink-next</id>
 			<properties>
-				<eclipselink>2.6.3-SNAPSHOT</eclipselink>
+				<eclipselink>2.6.5-SNAPSHOT</eclipselink>
```
olivergierke: Looks like we don't actually pull up the version of the version used for standard compilation to 2.6.4 then, do we? If we upgrade Eclipselink dependencies, let's upgrade them entirely, to the latest version available.

mp911de: I'll upgrade the regular dependency to 2.6.4 as well.

mp911de:
That's merged, polished and ported to maintenance branches.

PR #201, opened 2017-05-05
DATAJPA-1110 - Adapt to QueryByExampleExecutor API changes.

christophstrobl:
Use `Optional` as return type for `findOne(Example example)`.

See DATACMNS-1058 (spring-projects/spring-data-commons#214) for details.


mp911de:
That's merged now.

PR #206, opened 2017-07-07
DATAJPA-1140 - Allow mixing of SpEL expression and normal query parameters.

schauder:
Restructured the complete parameter binding.

There are now two principle ways to perform the binding:

If no query is known to the binder the method parameters get iterated and assigned to query parameters.
This is implemented by the bind method of the ParameterBinder.

If a query is known all the bindings, i.e the query parameters are iterated and matching values are searched.
This is implemented in the QueryAwareParameterBinder.

For finding a suitable value there exists various QueryParameterSetterStrategies.

The code calling the various Query.setParameter overloads is extracted and gathered in QueryParameterSetter.


src/main/java/org/springframework/data/jpa/repository/query/QueryAwareParameterBinder.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private final List<QueryParameterSetterStrategy> queryParameterSetterStrategies = new ArrayList<>();
+
+	/**
+	 * Creates a new {@link QueryAwareParameterBinder} from the given {@link Parameters}, method arguments and
+	 * {@link StringQuery}.
+	 * 
+	 * @param parameters must not be {@literal null}.
+	 * @param values must not be {@literal null}.
+	 * @param query must not be {@literal null}.
```
kevin-peters: should evaluationContextProvider & parser not be mentioned within javadoc?

src/main/java/org/springframework/data/jpa/repository/query/CriteriaQueryParameterBinder.java
```diff
@@ -60,19 +59,18 @@
 	 */
 	@Override
 	@SuppressWarnings("unchecked")
-	protected void bind(Query query, JpaParameter parameter, Object value, int position) {
+	protected QueryParameterSetter bind(JpaParameter parameter, Object value, Integer position) {
```
kevin-peters: The parameter "position" won't be used here? And why is it changed to Integer? It's only (at least within this context) invoked with primitive values, so it's boxed for nothing?

schauder: You are right on both accounts. Will fix once I got feedback from @olivergierke (assuming we are going forward with this approach).

olivergierke:
I guess we have to clarify on a couple of things as the PR implements quite a few very fundamental changes in terms of which component knows what. Not saying this is bad thing, I just want to make sure we understand what we're doing on a conceptual level.

Previously we had `ParameterBinder`, `StringQueryParameterBinder` and `SpelExpressionStringQueryParameterBinder`, which was by design to allow the middle one work without knowing anything about SpEL in the first place. With the introduction of `QueryAwareParameterBinder` this seems to have been gone now mixing SpEL stuff with non SpEL queries.

We now have `ParameterBinder`, `-Setter` and `-SetterStrategy` which we need to distinguish precisely: purpose, responsibilities and dependencies between them.

One other thing I'd like to have a second thought on is that we try to minimize object creation for actual query method invocations. The setter instances seem to be recreated for each query method invocation and parameter. I'd rather get to a object instance creation flow that can be statically created once per query method and the parameters being handed into the API methods then eventually bound using the reusable instance. I.e. abstractly speaking, on the very inner level, I'd like to get to this design:

```java
class SomeClass {

  private final Whatever whatever;

  public void bind(Object value) {
    whatever.someMethod(value);
  }
}
```

instead of:

```java
class SomeClass {

  private final Whatever whatever;
  private final Object value;

  public void bind() {
    whatever.someMethod(value);
  }
}
```

The latter requires reinstantiation for every `value`. In other words: the more moving a part is (with values changing all the time, the JPA query instance, too, the static string query and thus the binders not at all) the more they should be method parameters instead of class members.

Not saying we can avoid creating additional objects for parameters entirely, I just think we should try to avoid holding fields of different lifecycles in wrapping objects as the one with the shortest lifecycle defines the lifecycle for the wrapping class, too.

schauder:
Just checked, we can move the `value` from the constructor of the `ParameterSetters` to a parameter in a method. This way we can keep the `ParameterSetters` around.

PR #211, opened 2017-08-25
DATAJPA-1170 - Unify Specification and Specifications API.

koraktor:
Using default methods in the `Specification` interface we’re able to effectively remove the `Specifications` class from the public API.

This allows to use `and`, `or` and `not` without wrapping `Specification` into `Specifications` using `where`. `where` is still available to wrap potentially nullable specifications.


olivergierke:
That's polished and merged.

PR #217, opened 2017-09-08
DATAJPA-1172 (backport) Log when an exception gets swallowed during setting of parameters.

schauder:


FireWolf2007:
After upgrading to last modifications we have following exceptions:
```
2017-09-19 22:33:17.954  WARN 8 --- [ XNIO-2 task-54] SpelExpressionStringQueryParameterBinder : Setting the parameter with name 'null' and position '2' lead to an exception.

java.lang.IllegalArgumentException: Unknown parameter position: 2
	at org.hibernate.query.internal.QueryParameterBindingsImpl.getBinding(QueryParameterBindingsImpl.java:240)
```

Problem reproduces with select which has 2 identical SpEL expressions "?#{principal.username}"

Simplified part of JPQL causes this warning:
```
(EXISTS
	(
	SELECT 1 FROM User user LEFT JOIN user.authorities userAuthority
	WHERE user.login = ?#{principal.username} AND userAuthority.name IN('ADMIN','USER')
	)
	OR
 EXISTS
	(
	SELECT 1 FROM Employee employee INNER JOIN employee.userAccount user 
	WHERE user.login = ?#{principal.username} 
	)
)```

schauder:
This PR has uncovered a minor bug, which now triggers these warning.
There is a little more information in https://github.com/spring-projects/spring-data-examples/issues/293
There is already a [PR for a fix](https://github.com/spring-projects/spring-data-jpa/pull/220) for the [underlying bug](https://jira.spring.io/browse/DATAJPA-1179).

Other than the warnings in the logs there shouldn't be any negative effects for you or your application. And the logs should go away with an update in the near future.

PR #224, opened 2017-10-01
DATAJPA-1192 - Some JavaDocs and package-infos are not up-to-date

kevin-peters:
This PR corrects some JavaDocs which point to variables not existing (anymore) and other way around. It also fixes some broken @link references in the JavaDocs. At least it removes one a @NonNullApi annotation from package-info file where affected class methods may return null.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/convert/threeten/package-info.java
+++ b/src/main/java/org/springframework/data/jpa/convert/threeten/package-info.java
@@ -1,5 +1,4 @@
 /**
  * Spring Data JPA specific JSR-310 converters.
  */
-@org.springframework.lang.NonNullApi
 package org.springframework.data.jpa.convert.threeten;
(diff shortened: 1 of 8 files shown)


olivergierke:
Let's open a separate ticket for the commit regarding the nullable API. The justification for the commit is misleading as it's basically not an argument: we declare API non-nullable and then add `@Nullable` annotations to the methods that are an exception to that. Generally disabling that is not really an option. I'll go ahead and merge the other two commits.

kevin-peters:
Did create another ticket plus PR: [Add nullable (and override)](https://jira.spring.io/browse/DATAJPA-1194). Sorry, did not know that the @nonnullapi is mandatory even if all affected methods in (sub-)packages may return null. Will keep that in mind. ;)

schauder:
Seems this was already merged and just never got closed.

PR #225, opened 2017-10-02
DATAJPA-1194 - Add @Nullable annotations to methods in Jsr310JpaConverters

kevin-peters:
This PR adds @Nullable to methods which may return null. Also a missing @Override annotation is added.


olivergierke:
That's merged.

PR #227, opened 2017-10-16
DATAJPA-1205 - Reinstantiate entity manager injection in QueryDslRepositorySupport.

mp911de:
We now reinstantiated entity manager injection in QueryDslRepositorySupport using @AutoWired.

Previously, we removed dependency injection through @PersistenceContext to avoid bootstrap errors caused by class scanning in application containers signaled by deployment annotation use.

---

Related tickets: [DATAJPA-1205](https://jira.spring.io/browse/DATAJPA-1205), [DATAJPA-1175](https://jira.spring.io/browse/DATAJPA-1175)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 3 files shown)


izeye:
This looks merged already.

olivergierke:
Indeed, thanks!

PR #228, opened 2017-10-19
DATAJPA-1182 - Validate method parameters on query construction.

schauder:
For “In” and “NotIn” queries not providing a collection like argument results in an IllegalStateException at initialization time.
    
Conversely providing a collection as an argument for a query other than “In” or “NotIn” also throws such an exception.


schauder:
See Jira for review comments https://jira.spring.io/browse/DATAJPA-1182

mp911de:

ahaczewski:
Hi Guys,
What would be the best way of reporting bug against this PR? We have a situation where we have custom JPA `@Converter` for Protobuf's `ByteString`, which happens to be `Iterable`. The repository can't be created, because of this validation checking for `Iterable` where scalar is expected, but our `@Converter` works making scalar base-64 string from that `ByteString` and it worked perfectly with 2.1, but fails due to this validation in 2.2.

schauder:
@ahaczewski create an issue in Jira: https://jira.spring.io/projects/DATAJPA

ahaczewski:
Thanks @schauder, reported as https://jira.spring.io/browse/DATAJPA-1682, hope it is enough.

PR #231, opened 2017-10-25
DATAJPA-1200 - :: and \: no longer start a named parameter.

schauder:
Unified the query parsing in QueryUtils and StringQuery to use the same regex for matching named parameters.
Added look behind to not consider :: and \: start of a named parameter.
    
Unicode whitespace, control, formatting and punctuation characters except '.' and '_' are no longer considered legal parameter names.
    
Note: QueryUtils accepts # as start of a named parameter. StringQuery doesn't.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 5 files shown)


lukaseder:
Spring Data is parsing SQL with regular expressions? [Ghasp](https://stackoverflow.com/a/1732454/521799)

Does this help as an illustration:

```sql
SELECT 'abc :oops xyz' FROM dual
```

schauder:
I know, I know. I don't think anybody is proud about this.

lukaseder:
I know the feeling, having been there :)

PR #237, opened 2017-12-08
DATAJPA-931 - Avoid unnecessary merging on save.

schauder:
Checking if entity is already attached to entity manager before calling merge.
    
Fixed one test that was relying on the implicit flush triggered by the save.

See also: https://vladmihalcea.com/2016/07/19/jpa-persist-and-merge/

### DIFF
--- a/src/test/java/org/springframework/data/jpa/domain/support/AuditingEntityListenerTests.java
+++ b/src/test/java/org/springframework/data/jpa/domain/support/AuditingEntityListenerTests.java
@@ -41,6 +41,7 @@ import org.springframework.transaction.annotation.Transactional;
  * Integration test for {@link AuditingEntityListener}.
  *
  * @author Oliver Gierke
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/data/jpa/domain/support/AuditingEntityListenerTests.java
```diff
@@ -89,7 +90,7 @@ public void auditsTransitiveEntitiesCorrectly() {
 		role.setName("ADMIN");
 
 		user.addRole(role);
-		repository.save(user);
```
olivergierke: Why is this change needed? And if applied, shouldn't it be `saveAndFlush(…)`?

schauder:
merged: https://github.com/spring-projects/spring-data-jpa/commit/26f70bf1f002ac34f47a42bf06268f754ef0df6c

PR #244, opened 2018-01-23
DATAJPA-1235 - String literals that look like bind parameters now get ignored.

schauder:
Split in two commits to separate refactoring from change in functionality.

Refactoring: Move functionality into `StringQuery`

Functionality: Change the parser to honor String literals.


src/main/java/org/springframework/data/jpa/repository/query/QueryInformation.java
```diff
@@ -0,0 +1,132 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * A wrapper for a String representation of a query offering information about the query.
+ *
+ * @author Jens Schauder
+ */
+interface QueryInformation {
```
olivergierke: I'd prefer `DeclaredQuery` as it captures the value it captures slightly better.

src/main/java/org/springframework/data/jpa/repository/query/QueryInformation.java
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * An implementation implementing the NULL-Object pattern for situations where there is no query.
+	 */
+	QueryInformation EMPTY_QUERY = new EmptyQueryInformation();
+
+	static QueryInformation of(String query) {
```
olivergierke: JavaDoc?

src/main/java/org/springframework/data/jpa/repository/query/QueryInformation.java
```diff
@@ -0,0 +1,132 @@
...
+		return StringUtils.isEmpty(query) ? EMPTY_QUERY : new StringQuery(query);
+	}
+
+	/**
+	 * @return {@literal true} iff the underlying query has at least one named parameter.
```
olivergierke: Iff -> whether.

src/main/java/org/springframework/data/jpa/repository/query/QueryInformation.java
```diff
@@ -0,0 +1,132 @@
...
+	 * @return A new {@literal QueryInformation} instance.
+	 */
+	QueryInformation deriveCountQuery(@Nullable String countQuery, @Nullable String countQueryProjection);
+
+	/**
+	 * NULL-Object pattern implementation.
+	 *
+	 * @author Jens Schauder
+	 */
+	class EmptyQueryInformation implements QueryInformation {
```
olivergierke: Can we put this into a separate file as otherwise it's public by default? Rename to `EmptyDeclaredQuery`?

src/main/java/org/springframework/data/jpa/repository/query/QueryInformation.java
```diff
@@ -0,0 +1,132 @@
...
+ * A wrapper for a String representation of a query offering information about the query.
+ *
+ * @author Jens Schauder
+ */
+interface QueryInformation {
+
+	/**
+	 * An implementation implementing the NULL-Object pattern for situations where there is no query.
+	 */
+	QueryInformation EMPTY_QUERY = new EmptyQueryInformation();
```
olivergierke: Can we move this into `EmptyQueryInformation`, i.e. `EmptyQueryInformation.INSTANCE`? In combination with moving `EQI` to a separate file we can prevent the constant from leaking into public API.

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -201,10 +201,6 @@ public static String getExistsQueryString(String entityName, String countQueryPl
 
 	/**
 	 * Returns the query string for the given class name.
-	 *
```
olivergierke: Why remove the JavaDoc? Shouldn't we rather document the nullability here? Same for removals below.

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -77,14 +79,21 @@ boolean hasParameterBindings() {
 	/**
 	 * Returns the {@link ParameterBinding}s registered.
 	 */
-	List<ParameterBinding> getParameterBindings() {
```
olivergierke: Let's add `@Override` to the methods that are implementing the interface.

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -103,15 +112,29 @@ String getAlias() {
...
+		return getProjection().equals(alias);
+	}
+
+	public String getProjection() {
+		return QueryUtils.getProjection(query);
+	}
+
+	public boolean hasNamedParameter() {
+
+		for (ParameterBinding binding : bindings) {
```
olivergierke: `bindings.stream(…).anyMatch(…)`?

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -132,6 +155,7 @@ boolean isDefaultProjection() {
 		public static final int NAMED_PARAMETER_GROUP = 6;
 		public static final int COMPARISION_TYPE_GROUP = 1;
 		public static final int EXPRESSION_GROUP = 9;
+		public static final HashSet<Character> QUOTING_CHARACTERS = new HashSet<>(Arrays.asList('"', '\''));
```
olivergierke: `Set` on the assignment side?

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
schauder: Originally removed empty JavaDoc attributes to minimize noise.
Put those back in and added comments about nullability + annotations.

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -202,11 +203,12 @@ public static String getExistsQueryString(String entityName, String countQueryPl
...
 	 *
-	 * @param template
-	 * @param entityName
-	 * @return
+	 * @param template must not be {@literal null}.
+	 * @param entityName must not be {@literal null}.
+	 * @return the template with placeholders replaced by the {@literal entityName}. Guaranteed to be not {@literal null}.
 	 */
-	public static String getQueryString(String template, String entityName) {
+	@NonNull
```
olivergierke: Why do we need explicit `@NonNull` annotations here? Isn't that the default anyway?

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -163,6 +189,13 @@ boolean isDefaultProjection() {
 			PARAMETER_BINDING_PATTERN = Pattern.compile(builder.toString(), CASE_INSENSITIVE);
 		}
 
+		static List<ParameterFinds> parseParameters(String query) {
+			return new ArrayList<>();
+
+		}
+
+		private static class ParameterFinds {}
```
olivergierke: Can this be removed?

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -132,6 +157,7 @@ boolean isDefaultProjection() {
 		public static final int NAMED_PARAMETER_GROUP = 6;
 		public static final int COMPARISION_TYPE_GROUP = 1;
 		public static final int EXPRESSION_GROUP = 9;
+		public static final Set<Character> QUOTING_CHARACTERS = new HashSet<>(Arrays.asList('"', '\''));
```
olivergierke: Should we move that into `QuotationMap`?

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -163,6 +189,13 @@ boolean isDefaultProjection() {
 			PARAMETER_BINDING_PATTERN = Pattern.compile(builder.toString(), CASE_INSENSITIVE);
 		}
 
+		static List<ParameterFinds> parseParameters(String query) {
```
olivergierke: Is that intended? Ignoring the parameters looks weird.

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
schauder: Wasn't aware that `@NonNull` is the default, I thought that was controlled by a package level annotation. Removing `@NonNull`s again.

src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
schauder: That was a leftover from a different approach. :face_palm:  Removed it.

PR #245, opened 2018-01-26
 DATAJPA-1248 - Added build profiles for Hibernate 5.3.

schauder:
Removed obsolete version detection and associated persistence units from tests.

Added explicite cascade option to many-to-one reference in id.
It is not clear if this is a bug in Hibernate or not.
See HHH-12251 for details.

The fix of HHH-12119 made it obvious that our use of Tuples for Projections from native queries were not correct.
When looking up Tuple values we now find them using a property name with correct upper/lower case as well as with the upper case as it is actually returned by the JDBC driver.

Converted all touched tests using Hamcrest to use AssertJ instead.

See also:
https://hibernate.atlassian.net/browse/HHH-12251
https://hibernate.atlassian.net/browse/HHH-12119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ env:
     - PROFILE=ci
     - PROFILE=spring5-next
     - PROFILE=hibernate-next
+    - PROFILE=hibernate-53
+    - PROFILE=hibernate-53-next
     - PROFILE=eclipselink-next
     - PROFILE=eclipselink-27-next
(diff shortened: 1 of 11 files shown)


izeye:
I guess this can be closed now as [the mentioned issue](https://jira.spring.io/browse/DATAJPA-1248) has been already resolved and the commits belonging to this PR look merged to all active branches (`master`, `1.11.x`, `2.0.x`).

PR #246, opened 2018-01-29
 DATAJPA-928 - Enabled native query with Pageable.

schauder:
Just removed the check that was actively preventing the use of Pageable.

Migrated to tests to AssertJ where applicable.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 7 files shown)


izeye:
This looks resolved.

alexformagio:
Good job. Only one question. When do you think this change will be available? Thanks in advance.

olivergierke:
The referred to ticket lists Kay SR4 and Lovelace M1 as fixed versions, which means it's already shipped. Current Spring Boot 2.0 already contains the fix (as it's on Kay SR5).

alexformagio:
Great, Thanks for your reply!

PR #249, opened 2018-02-05
 DATAJPA-1261 - Revert DATAJPA-931.

schauder:
This reverts the functional change made by DATAJPA-931.
This means merge will be called again, even if the entity for which save was called is already attached to the session.
Tests stay in place to verify the new old behavior.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


stsypanov:
Could anyone explain in brief why this change was reverted in 2.0.X? Any ticket?

madness-inc:
The original ticket an be found here:
https://jira.spring.io/browse/DATAJPA-931
This is the related commit:
https://github.com/spring-projects/spring-data-jpa/commit/8265daffa00fe19e73d6f7039a288ad485b8751a

But I agree that it was not stated why it was reverted.

stsypanov:
Thanks, I'll ask this question there.

PR #250, opened 2018-02-21
DATAJPA-1267 - Using SpEL infrastructure from commons.

schauder:
This PR depends on DATACMNS-1258 (https://github.com/spring-projects/spring-data-commons/pull/275) which needs to get merged first.

With DATACMNS-1258 handling of SpEL expressions in queries was moved to SD-Commons.
This change now uses that infrastructure and removes the parts no longer necessary in this repository.


mp911de:

PR #251, opened 2018-02-22
DATAJPA-1105 - Fixed support for nested IdClass with non mapped non primitive type.

schauder:
When obtaining a multipart id from an entity we have to check for each part of the id if it is another entity.

Before this change that check was done by checking if the type is a primitive or wrapper type, which fails to properly classify e.g. String.
This caused us to try to get an id from a String instance which of course failed because String is not a managed type.

Now we make the necessary distinction based on JpaMetamodel.isJpaManaged(type).

Changes in the persistence.xml beyond adding entities for tests are required to make the tests work again with EclipseLink.
This didn't cause problems in the past because all tests that actually access the database and use the changed persistence context are disabled for EclipseLink.

The new test had to get disabled for EclipseLink though, due to another bug in EclipseLink which prevents the usage of inner classes in @IdClass annotations.

See also: https://bugs.eclipse.org/bugs/show_bug.cgi?id=531528

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
+	<version>2.1.0.DATAJPA-1105-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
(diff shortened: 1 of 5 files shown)


izeye:
This looks merged into `1.11.x`, `2.0.x` and `master` branches.

olivergierke:
Indeed, it is. Closing.

PR #255, opened 2018-03-05
DATAJPA-1281 - Provide a better error message on mismatch between required and actual parameters.

schauder:
The added assertion prevents the non-descriptive `IndexOutOfBoundsException` and provides a more digestible message.


mp911de:

PR #260, opened 2018-03-22
DATAJPA-1297 - Fixed usage of outputParameterName in Procedure annot…

didiez:
…ation.


schauder:
That looks already quite nice. 
One big improvement of the PR would be to have some integration test.

schauder:
Thanks, that's merged.

PR #261, opened 2018-03-24
DATAJPA-1303. Added support ignoring case for finding collections

krupt:

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
@@ -193,7 +193,7 @@ class ParameterMetadataProvider {
 		private final Type type;
 		private final ParameterExpression<T> expression;
 		private final PersistenceProvider persistenceProvider;
-		private final Part.IgnoreCaseType ignoreCaseType;
+		private final boolean ignoreCase;
 
 		/**
 		 * Creates a new {@link ParameterMetadata}.
@@ -204,7 +204,7 @@ class ParameterMetadataProvider {
 			this.expression = expression;
 			this.persistenceProvider = provider;
 			this.type = value == null && Type.SIMPLE_PROPERTY.equals(part.getType()) ? Type.IS_NULL : part.getType();
-			this.ignoreCaseType = part.shouldIgnoreCase();
+			this.ignoreCase = IgnoreCaseType.ALWAYS.equals(part.shouldIgnoreCase());
 		}
 
 		/**
@@ -251,7 +251,7 @@ class ParameterMetadataProvider {
 			}
 
 			return Collection.class.isAssignableFrom(expressionType) //
-					? persistenceProvider.potentiallyConvertEmptyCollection(upperIfIgnoreCase(ignoreCaseType, toCollection(value))) //
+					? persistenceProvider.potentiallyConvertEmptyCollection(upperIfIgnoreCase(ignoreCase, toCollection(value))) //
 					: value;
 		}
 
@@ -283,8 +283,9 @@ class ParameterMetadataProvider {
 
 		@Nullable
 		@SuppressWarnings("unchecked")
-		private static Collection<?> upperIfIgnoreCase(Part.IgnoreCaseType ignoreCaseType, @Nullable Collection<?> collection) {
-			if (IgnoreCaseType.ALWAYS.equals(ignoreCaseType)
+		private static Collection<?> upperIfIgnoreCase(boolean ignoreCase, @Nullable Collection<?>
+				collection) {
+			if (ignoreCase
 					&& !CollectionUtils.isEmpty(collection)) {
 				return ((Collection<String>) collection).stream()
 					.map(it -> {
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@krupt Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=261)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=261) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@krupt Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=261)!

src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
```diff
@@ -185,19 +188,21 @@ private ParameterMetadataProvider(CriteriaBuilder builder, @Nullable Iterator<Ob
...
 
+		private final Part part;
 		private final Type type;
 		private final ParameterExpression<T> expression;
 		private final PersistenceProvider persistenceProvider;
 
 		/**
 		 * Creates a new {@link ParameterMetadata}.
 		 */
-		public ParameterMetadata(ParameterExpression<T> expression, Type type, @Nullable Object value,
```
olivergierke: Can you clarify why this change is needed? I don't feel comfortable with making `ParameterMetadata` aware of anything ignore-case related. Shouldn't the handling of that be applied in the code eventually calling `….prepare(…)`?

src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
```diff
@@ -232,19 +237,19 @@ public Object prepare(Object value) {
 
 				switch (type) {
 					case STARTING_WITH:
-						return String.format("%s%%", value.toString());
```
olivergierke: Please avoid changes unrelated to the actual improvement we're trying to fix.

src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
```diff
@@ -232,19 +237,19 @@ public Object prepare(Object value) {
...
 					case CONTAINING:
 					case NOT_CONTAINING:
-						return String.format("%%%s%%", value.toString());
+						return "%" + value + "%";
 					default:
 						return value;
 				}
 			}
 
-			return Collection.class.isAssignableFrom(expressionType) //
```
olivergierke: Please avoid changes unrelated to the actual improvement we're trying to fix.

src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
```diff
@@ -1175,6 +1175,48 @@ public void findByElementCollectionAttribute() {
 		assertThat(result).containsOnly(firstUser, secondUser);
 	}
 
+	@Test
```
olivergierke: Please add references to the JIRA issue like the other test methods do. Also, please add new test methods *last*, not in the middle of others.

src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java
```diff
@@ -290,6 +290,12 @@
 	// DATAJPA-496
 	List<User> findByAttributesIn(Set<String> attributes);
 
+	List<User> findByAttributesIgnoreCaseIn(Collection<String> attributes);
```
olivergierke: Please move those methods to the bottom of the file and mark them with the JIRA ticket because of which they were introduced.

src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
krupt: Should I create another issue in JIRA to fix creation of unnecessary objects?

olivergierke: Unless you have detailed profiling numbers showing that that change has significant impact on the performance I don't think it's worth the effort.

src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
krupt: I need more time to study this proposal

krupt: I can't find better solution. Method .prepare calling only from org.springframework.data.jpa.repository.query.QueryParameterSetterFactory.CriteriaQueryParameterSetterFactory#CriteriaQueryParameterSetterFactory, but there have not place for this. In ParameterMetadata we prepare another string conversion and can do this conversion. I think that I should change ignoreCaseType field to boolean

schauder:
Thanks, that is squashed, polished and merged.

PR #270, opened 2018-04-24
DATAJPA-1238 Unnecessary LEFT OUTER JOIN when a WHERE could have been used

reda-alaoui:


src/test/java/org/springframework/data/jpa/repository/query/QueryUtilsIntegrationTests.java
```diff
@@ -87,7 +87,7 @@ public void createsJoinForOptionalAssociation() {
 		CriteriaQuery<User> query = builder.createQuery(User.class);
 		Root<User> root = query.from(User.class);
 
-		QueryUtils.toExpressionRecursively(root, PropertyPath.from("manager", User.class));
+		QueryUtils.toExpressionRecursively(root, PropertyPath.from("manager.firstname", User.class));
```
schauder: Please don't change existing tests if their behavior isn't supposed to change. Add a new one instead.

The real problem is that the test doesn't fail without your changes, so it doesn't really test the changes. 
Also, I'm thinking you might be working on a different issue than the original Jira ticket. 
I'll comment about that on the ticket.

schauder: Ok, misinterpreted your change as a new test. 

It's more, that the new test is missing. I'll add a test, and merge.

schauder:
Thanks for the work. That's polished and merged.

PR #273, opened 2018-04-27
DATAJPA-1336 remove unused imports

Chr1st0ph:
DATAJPA-1336 remove unused imports


schauder:
That's merged. Thank you.

PR #274, opened 2018-05-03
Added a cover page for epub output

Buzzardo:
I added a cover page for the epub output, now that the Spring data reference guides will
have epub output (through a commit against spring-data-build). I also added my name to the author list, added .DS_Store to .gitignore, and caught a couple other things.


mp911de:

PR #277, opened 2018-05-18
DATAJPA-1350 - Add utility to use JPA in a reactive environment.

mp911de:
Provide an Executor API that allows `EntityManager` and `JpaRepository` usage within a reactive application environment. Workloads are offloaded to a dedicated Scheduler to unblock the subscribing thread and to not block application threads. `DefaultJpaExecutor` requires a dedicated `Scheduler` that is exclusively used to process blocking workloads.

```java
Flux<User> read = executor.readMany(repository, JpaRepository::findAll);

Flux<User> read = executor.readMany(em -> {
	return em.createQuery("SELECT u FROM User u WHERE u.firstname = ?0")
				.setParameter(0, "foo")
				.getResultList();
});

Mono<Void> delete = executor.transactional().doInTransaction(repository, userRepository -> {
	userRepository.deleteAll();
	return null;
});

Mono<Void> delete = executor.transactional() //
		.withNoRollbackOn(RuntimeException.class) //
		.withRollbackOn(IllegalStateException.class) //
		.doInTransaction(repository, userRepository -> {
			userRepository.deleteAll();
			throw new IllegalStateException("e");
});
```

---

Related ticket: [DATAJPA-1350](https://jira.spring.io/browse/DATAJPA-1350).
See also https://github.com/spring-projects/spring-data-examples/tree/3fa8586b398f756994fb962ca6cb6fb82cf50e7d/web/webflux to see usage of `JpaExecutor` in the context of a WebFlux application.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
+	<version>2.1.0.DATAJPA-1350-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
 	<description>Spring Data module for JPA repositories.</description>
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,18 @@
 		</dependency>
 
 		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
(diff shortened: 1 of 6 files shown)


mp911de:
We're not merging this pull request yet as we're investigating on reactive relational database access.

mp911de:
With Spring Data R2DBC, there is no longer a need to build a bridge as we do not recommend offloading blocking calls to thread pools.

PR #278, opened 2018-06-05
DATAJPA-1359 - Remove reference to org.hibernate.Query by simplifying HibernateUtils.

schauder:
HibernateUtils made heavy use of reflection in order to support different versions of Hibernate.
With the current required version of Hibernate as indicated by the pom.xml this degree of flexibility is no longer needed.

Also:
Upgraded the Hibernate dependencies because 5.3.1 is now final and the latest Snapshot seems to be broken.
Therefore it seems sensible to depend on 5.3.1-Final and 5.3.2-SNAPSHOT respectively in hibernate-53 and hibernate-53-next.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,13 +48,13 @@
 		<profile>
 			<id>hibernate-53</id>
 			<properties>
-				<hibernate>5.3.0.Final</hibernate>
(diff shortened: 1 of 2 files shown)


schauder:
This is already merged.

PR #279, opened 2018-06-06
 DATAJPA-1156 Filter QueryHints for count queries in QuerydslJpaPredicateExecutor.

schauder:


mp911de:

PR #280, opened 2018-07-03
DATAJPA-1368 - Register ExposeRepositoryInvocationInterceptor to isolate repository method call exposure.

mp911de:
We now ship and register `ExposeRepositoryInvocationInterceptor` to expose proxy method invocations without interfering with `ExposeInvocationInterceptor`. `ExposeRepositoryInvocationInterceptor` is a copy of `ExposeInvocationInterceptor` that holds its own instance of the invoked method and that is isolated so no other component can affect the exposure of the invoked method.

Interceptors in the invocation chain can invoke methods on proxied objects (e.g. proxied `TransactionManager`) which override the invoked proxied method as `ExposeInvocationInterceptor` is a singleton. We previously lost the invoked method details.

---

Related ticket: [DATAJPA-1368](https://jira.spring.io/browse/DATAJPA-1368).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


schauder:
Done.

PR #281, opened 2018-07-09
DATAJPA-1372 - Allow Query by Example matching for joined singular attributes

mp911de:
We now support Query by Example matching for singular attributes that are joined through embeddable entities. Previously, we rejected these as the metamodel returns a `SingularAttributeJoin` whereas we always expect a `From` to create a join.

cc @christophstrobl 

---

Related ticket: [DATAJPA-1372](https://jira.spring.io/browse/DATAJPA-1372).


schauder:
Merged.

PR #282, opened 2018-07-09
Adding @EntityGraph example using custom annotation

marceloverdijk:
Adding @EntityGraph example using custom annotation, see also https://jira.spring.io/browse/DATAJPA-1373


schauder:
The use of meta-annotations is a general concept, that doesn't just apply to `EntityGraph`. So if we want to add documentation about that we should do it in the general part of the reference documentation, i.e. in Spring Data Commons.

PR #283, opened 2018-07-13
DATAJPA-1163 - Allow expression "#{#entityName}" in count queries

thombergs:
This PR solves [DATAJPA-1163](https://jira.spring.io/browse/DATAJPA-1163) and allows using the expression "#{#entityName}" in count queries.

The issue described in DATAJPA-1163 was reproducible with the test case I added.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
@@ -34,6 +34,7 @@ import org.springframework.util.Assert;
  * @author Oliver Gierke
  * @author Thomas Darimont
  * @author Jens Schauder
+ * @author Tom Hombergs
  */
 abstract class AbstractStringBasedJpaQuery extends AbstractJpaQuery {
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/jpa/repository/query/DeclaredQuery.java
```diff
@@ -80,9 +82,12 @@ static DeclaredQuery of(@Nullable String query) {
 	 * 
 	 * @param countQuery an optional query string to be used if present.
 	 * @param countQueryProjection an optional return type for the query.
+	 * @param metadata metadata about the queried entity.
+	 * @param parser SpEL parser for evaluating expressions within the query.
 	 * @return a new {@literal DeclaredQuery} instance.
 	 */
-	DeclaredQuery deriveCountQuery(@Nullable String countQuery, @Nullable String countQueryProjection);
+	DeclaredQuery deriveCountQuery(@Nullable String countQuery, @Nullable String countQueryProjection,
```
schauder: Adding the parser as an argument here makes this rather basic class dependent on infrastructure which is really only needed for `ExpressionBasedStringQuery`. I'd therefor rather leave this method unchanged and instead add a factory method in `ExpressionBasedStringQuery` which takes a `DeclaredQuery` and turns it into an `ExpressionBasedStringQuery`. That method can then be used in `AbstractStringBasedJpaQuery`.

src/test/java/org/springframework/data/jpa/repository/query/NamedQueryUnitTests.java
```diff
@@ -15,7 +15,8 @@
  */
 package org.springframework.data.jpa.repository.query;
 
-import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.ArgumentMatchers.any;
```
schauder: Please make sure that static imports stay as wildcard imports.

thombergs: No longer relevant in the new push since this file wasn't touched.

schauder:
Thanks, that's polished and merged.

PR #286, opened 2018-07-27
DATAJPA-1384 - Only entities and embeddables are considered mapped types.

schauder:
Other mapped types (according to Hibernate) are not relevant to us and trigger wrong behavior like not requesting a Tuple result when the requested result type is Map and by using Envers Map is considered a mapped type by Hibernate.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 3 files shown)


olivergierke:
We seem to single out mapped supertypes now and why we do that is not immediately obvious. Even more so, I'm not convinced that this is correct as we exclude those in a method that explicitly asks for *all* managed types, which is what mapped superclasses actually are.

schauder:
Hibernate actually fixed this on their side. https://hibernate.atlassian.net/browse/HHH-12871

PR #287, opened 2018-07-31
DATAJPA-1379 - Fixed detection of PersistenceProvider when EntityManager is a proxy.

schauder:
We now use the delegate to test the applicability of the PersistenceProvider not just for determining the and the key cash.

This fixes DATAJPA-1379 because the underlying issue was that the generic PersistenceProvider was used.
Which return false for `canExtractQuery` which prevented the constructor call in the query to be detected.


mp911de:

PR #292, opened 2018-09-03
DATAJPA-1415 Specification totalElements's Bug ,When we use group By and pageable

yangcong3643:
When we use both group By and pageable,for example:

Page<User> page = repository.findAll((root, query, cb) ->

{ query.groupBy(root.get("clazz")); return null; }
, PageRequest.of(0,10));

the page's total is not the total num after grouping,unless it's the last page.

I see in the SimpleJpaRepository.executeCountQuery, it add the count every group, but I only need the size of group,so I changed it.

I'm sorry I didn't create a test case, I'm not good with Mockito framework, and it says don't mock value objects.I don't know what to do.


schauder:
I'm closing this issue since the underlying issue is not supported. See https://jira.spring.io/browse/DATAJPA-1415 for more information.

PR #294, opened 2018-10-02
DATAJPA-1418 Interface-based Projections - Generate inner join instead of left join

reda-alaoui:
If the PR is accepted, I think a merge to 2.0.x is needed.

By default, when a foreign entity, without further sub attribute, is involved, Hibernate generates an inner join on the table of the foreign entity. Hibernate internals do this before processing the EntityGraph. So even when the final SQL query only fetch the foreign entity ID hold by the primary table, it joins on the foreign table. I don't think Hibernate is going to change this behaviour before 6.0.

The current bug was introduced by DATAJPA-1238 fix. The goal of DATAJPA-1238 was to avoid a useless left outer join when a foreign entity was involved in the query predicate (findByX where X is an entity). Its fix removed the outer join when the navigated foreign entity was a leaf, regardless of the location of the property representing the entity.

My pull request creates a distinction between properties navigated inside the select clause and other properties (i.e. navigated in the predicate). If the entity property is a leaf outside a select clause, DATAJPA-1238 fix is kept, no outer join is generated. If the property is a leaf inside a select clause, an outer join is created as it was before DATAJPA-1238 fix.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryCreator.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryCreator.java
@@ -52,6 +52,7 @@ import org.springframework.util.Assert;
  * @author Mark Paluch
  * @author Michael Cramer
  * @author Mark Paluch
+ * @author Reda.Housni-Alaoui
  */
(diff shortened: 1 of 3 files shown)


schauder:
Thanks, that is merged.

reda-alaoui:
Thanks :)

PR #298, opened 2018-10-22
DATAJPA-1449 - Remove reference to `Specifications` in docs

koraktor:
`Specifications` has been deprecated since 2.0, so it shouldn’t be mentioned in the docs. Instead, refer to the default methods of `Specification`.


schauder:
That's merged and backported. Thanks.

PR #300, opened 2018-10-25
DATAJPA-1449 - Prepare removal of `Specifications`

koraktor:
`Specifications` is deprecated since 2.0. This patch prepares the code for removing it completely.

 * Moves static helper code into a private helper class
 * Removes references to `Specifications` from `Specification`

This should make it pretty straight forward to remove the class completely (in 3.0?) without affecting compatibility right now.


odrotbohm:
Great suggestion! I think we can even go ahead and remove `Specifications` for 2.2 as we've deprecated it in 2.0 already. Would you mind adapting your PR accordingly?

koraktor:
No problem, done.

schauder:
That's squashed and merged. Thanks.

PR #301, opened 2018-10-25
DATAJPA-1446 - Clear JpaMetamodel CACHE when application context is c…

Nowheresly:
…losed.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/config/JpaMetamodelMappingContextFactoryBean.java
+++ b/src/main/java/org/springframework/data/jpa/repository/config/JpaMetamodelMappingContextFactoryBean.java
@@ -31,6 +31,7 @@ import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.data.jpa.mapping.JpaMetamodelMappingContext;
+import org.springframework.data.jpa.util.JpaMetamodel;
 import org.springframework.data.util.StreamUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/jpa/mapping/JpaMetamodelMappingContext.java
```diff
@@ -109,6 +112,11 @@ public boolean hasPersistentEntityFor(Class<?> type) {
 		return super.hasPersistentEntityFor(type) || models.isMetamodelManagedType(type);
 	}
 
+	@EventListener
```
dsyer: Spring library components are not supposed to use `@EventListener` (it's a choice for application developers, but it's really inefficient, so not a good idea for libraries). You could put this in a `@PreDestroy` though, I expect (or `DisposableBean` probably better).

Nowheresly: Thanks for the review! I just updated my PR to use a `DisposableBean` instead.

Nowheresly: Well, implementing `DisposableBean` in class `JpaMetamodelMappingContext` was not triggering anything when application context is closed. Since `JpaMetamodelMappingContextFactoryBean` already implements `DisposableBean`, I decided to update my PR to clean the cache here instead.
This time, the cache is cleaned when applicationContext is closed.

odrotbohm:
That's complete. I took a slightly different approach, see the ticket comment for details. Kept the credits in the classes around and attributed you in the commit message.

PR #303, opened 2018-11-18
DATAJPA-1451 - Inconsistent Assert messages.

moose56:
Make Assert.notNull and Assert.notEmpty messages and usage consistent in the project.


schauder:
That is squashed, rebased and backported. 

Thanks.

PR #304, opened 2018-12-19
DATAJPA-1404 - Create explicit outer join for inverse, optional one-to-one relationships.

schauder:
When deciding between simple navigation and explicit outer join we consider this additional case to work around peculiar behavior of JPA implementations.

See also:
https://hibernate.atlassian.net/browse/HHH-12712
eclipse-ee4j/jpa-api#170


mp911de:

PR #375, opened 2019-04-24
 DATAJPA-1533 - Fix the Byte Buddy version to a Mockito compatible one.

schauder:
See also: https://github.com/mockito/mockito/issues/1606


mp911de:

PR #376, opened 2019-04-24
 DATAJPA-1530 - Properly support new Hibernate proxies.

schauder:
I didn't find a reasonable way to test this :-/


schauder:
Merge only after DATAJPA-1533. Without that the hibernate-53x branches are broken anyway.

mp911de:

PR #378, opened 2019-05-02
 DATAJPA-1535 - When deleting a new entity we no longer merge it.

schauder:
The merge happened to allow deleting of unmanaged entities.
As a side effect it triggered an insert for new (and thereby also unmanaged) entities.
By checking if the entity is new we avoid this now.


mp911de:

PR #379, opened 2019-05-09
DATAJPA-1506 - Fix alias detection for native query.

florianluediger:
When no alias is specified in a native query that does not contain a WHERE clause right after the FROM clause, this is now correctly detected.

[Link to Jira Ticket](https://jira.spring.io/browse/DATAJPA-1506)

I have added support for the `ORDER BY` and `GROUP BY` clauses because these are fairly common and available in almost every SQL flavor. Should I consider to add more possible clauses like `LIMIT`?


odrotbohm:
Merged, polished and backported. Thanks!

PR #380, opened 2019-05-17
DATAJPA-1500 - Fix RegEx to support whitespace characters at the end.

florianluediger:
When using the creteCountQueryFor method before, the order by clause did not get removed when having a specific combination of whitespace characters at the end of the input.
By removing the $ for matching the end of the line, this is now fixed.

[Link to the issue in Jira](https://jira.spring.io/browse/DATAJPA-1500)


schauder:
Thanks. 
I applied a tiny bit of formatting, merged and backported it.

jnizet:
Thanks for the fix @florianluediger and @schauder.

PR #381, opened 2019-05-31
DATAJPA-1435 - change COMPLEX_COUNT_VALUE for native query pageable sql.

JIANGc:

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
@@ -36,6 +36,7 @@ import org.springframework.util.Assert;
  * @author Jens Schauder
  * @author Tom Hombergs
  * @author David Madden
+ * @author Chao Jiang
  */
 abstract class AbstractStringBasedJpaQuery extends AbstractJpaQuery {
 
@@ -66,7 +67,7 @@ abstract class AbstractStringBasedJpaQuery extends AbstractJpaQuery {
 		this.evaluationContextProvider = evaluationContextProvider;
 		this.query = new ExpressionBasedStringQuery(queryString, method.getEntityInformation(), parser);
 
-		DeclaredQuery countQuery = query.deriveCountQuery(method.getCountQuery(), method.getCountQueryProjection());
+		DeclaredQuery countQuery = query.deriveCountQuery(method.getCountQuery(), method.getCountQueryProjection(), method.isNativeQuery());
 		this.countQuery = ExpressionBasedStringQuery.from(countQuery, method.getEntityInformation(), parser);
 
 		this.parser = parser;
--- a/src/main/java/org/springframework/data/jpa/repository/query/DeclaredQuery.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/DeclaredQuery.java
@@ -24,6 +24,7 @@ import org.springframework.util.StringUtils;
  * A wrapper for a String representation of a query offering information about the query.
  *
  * @author Jens Schauder
+ * @author Chao Jiang
  * @since 2.0.3
  */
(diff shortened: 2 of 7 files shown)


pivotal-issuemaster:
@JIANGc Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=381)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=381) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@JIANGc Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=381)!

JIANGc:
A native query ex. 

SELECT a.id AS id, a.time AS createTime, a.item AS item, a.category AS category, a.biz AS business, a.c_type AS contentType, a.cp_api AS cpApi, a.dedup AS dedup, a.content_level AS contentLevel, a.reserve_cp_api AS reserveCpApi, SUM(b.value) AS dedupCount, SUM(c.value) AS originalCount FROM statis_data AS a LEFT JOIN statis_data AS b ON a.id = b.id AND b.dedup = 1 LEFT JOIN statis_data AS c ON a.id = c.id AND c.dedup = 0 WHERE (COALESCE(:contentLevel) is NULL OR a.content_level IN (:contentLevel)) AND (COALESCE(:contentType) is NULL OR a.c_type IN (:contentType)) AND (COALESCE(:contentAppId) is NULL OR a.biz IN (:contentAppId)) AND (COALESCE(:cpApi) is NULL OR a.cp_api IN (:cpApi)) AND a.category = 'total' AND a.time BETWEEN :startTime AND :endTime AND a.item = 'content_accept_num' GROUP BY a.cp_api, a.c_type, a.content_level, a.biz

then after countQuery = matcher.replaceFirst(String.format(COUNT_REPLACEMENT_TEMPLATE, replacement));

countQuery = "select count(a) FROM  statis_data AS a LEFT JOIN statis_data AS b ON a.id = b.id AND b.d ... ..."

$3$6 in regex get the table name

JIANGc:
@mp911de any hope for this one getting merged? Thank you for the review.

schauder:
I'll try to look into this next week.
Since the change is surprisingly small chances aren't to bad for it.
The challenges will be:
- does it actually fix the problem? This is currently hard to judge because there is no testcase reproducing the original problem. Although the description of the issue might be detailed enough to create one.
- I'll have to think hard if it might break anything.

JIANGc:
I found that this code will have problem if there is distinct in sql.

Another solution is that we set 	
private static final String COMPLEX_COUNT_VALUE = "$3$4";

and use sub queries in From clause. In  this case we can keep distinct, all functions and all column names.

schauder:
I like the subquery approach. Sounds less brittle than the current "Take that SQL statement".
The problem with that though is that JPQL doesn't support subqueries in the FROM-clause as far as I know and interpret the specification.

An actual test case would be very helpfull, because right now I don't even know what the exact problem is.

JIANGc:
Thanks @schauder. Indeed. I submitted an update. JPQL and SQL can be parsed separately now.
I added also some test cases. Hope this can be helpful.

schauder:
I refactored the changes because they broke backward compatibility. 
And I also found a case where this breaks: an order by clause doesn't get properly removed.

The refactored version including a failing test is on  branch pr/datajpa-1435

JIANGc:
Indeed. The regex remove all characters after order by clause. I made some changes to fix this. I also create another pull request(DATAJPA-1435 - fix order by clause problem) to branch pr/datajpa-1435.

schauder:
Superseded by #385.

PR #384, opened 2019-06-18
DATAJPA-1560 - Avoid unnecessary creation of exception when checking if a type is managed.

schauder:
We check a type is managed before trying to access its `ManagedType`.

https://jira.spring.io/browse/DATAJPA-1560

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
+	<version>2.2.0.DATAJPA-1560-SNAPSHOT</version>
 
 	<name>Spring Data JPA</name>
 	<description>Spring Data module for JPA repositories.</description>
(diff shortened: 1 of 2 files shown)


shahamit:
When is this fix planned to be merged in master? If it's going to take a few days, can someone share the location (url) from where we can download the '2.2.0.BUILD-SNAPSHOT' jar so that we can validate the cpu issue locally?

schauder:
It will take a couple of days.

Snapshots get published here:
https://repo.spring.io/snapshot/

This particular one is here: 
https://repo.spring.io/snapshot/org/springframework/data/spring-data-jpa/2.2.0.DATAJPA-1560-SNAPSHOT/

shahamit:
@schauder - I understand that this is not the right place to ask but where can I find the release dates for upcoming spring-data-jpa? I am looking forward for this fix but I cannot upgrade to milestone releases. I could not find a page that shows the dates for 2.2 release?

schauder:
@shahamit we don't advertise release dates in advance because it just leads to frustration and anger when a date slips.

That said it is probably save to assume that a GA release will ship before the Spring One Platform conference at the beginning of October

schauder:
This shouldn't have been closed.

mp911de:
That's merged, polished, and backported.

PR #386, opened 2019-06-21
DATAJPA-1564 - Fix RegEx to support new lines in select clause using embedded DOTALL…

mhewedy:
It is now very common to use the multiline string in Kotlin to compose a data jpa query, especially if the query is complex and not as simple as the one in the example below:

```kotlin
@Query("""
        select o.id,
        o.name,
        o.age
        from User o
        where o.age = :age
    """)
    fun findUserByAge(@Param("age") age: Int, pageable: Pageable): Page<User>
```
currently, the countQuery will not be created successfully. (check the test case)


schauder:
@mhewedy please [create an issue in Jira](https://jira.spring.io/projects/DATAJPA) explaining the current behaviour and why it is wrong. 

Especially what you mean by 

> the countQuery will not be created successfully

needs some clarification

schauder:
From looking at the test you provided, this might be a duplicate of https://jira.spring.io/browse/DATAJPA-1500

mhewedy:
I didn't create a JIRA issue at first place cause I knew that spring fully migrated from JIRA. Issue created https://jira.spring.io/browse/DATAJPA-1564

> From looking at the test you provided, this might be a duplicate of https://jira.spring.io/browse/DATAJPA-1500

Not it is not. I've checked DATAJPA-1500 before sending the PR.

To understand what I mean, better than me to explain, you can just run this test case on the version of the code:

```java
@Test
public void createCountQuerySupportsNewLinesInSelectClause() {

	assertThat(createCountQueryFor("select user.age,\n" + //
					"  user.name\n" + //
					"  from User user\n" + //
					"  where user.age = 18\n" + //
					"  order by user.name\n "), //
			is("select count(user) from User user\n" + //
					"  where user.age = 18\n "));
}
```

schauder:
Thanks for the Jira ticket. 

Not all of Spring moved to Github issues. One little subproject still is on Jira. Spring Data.

schauder:
This still breaks when the linebreak is right after a distinct:

```
		assertThat(createCountQueryFor("select\ndistinct\nuser.age,\n" + //
						"user.name\n" + //
						"from\nUser\nuser"), //
				is(createCountQueryFor("select\ndistinct user.age,\n" + //
						"user.name\n" + //
						"from\nUser\nuser")));
```
and probably also when the line break is before the `BY` of `order by` or `group by`

Could you take care of that as well?

mhewedy:
Yea sure

mhewedy:
PR updated to support line breaks right after distinct.

I've used https://www.freeformatter.com/java-regex-tester.html for dissecting the count regex.

mhewedy:
I'll give the order/group by a look soon.

mhewedy:
I've updated the `ALIAS_MATCH` regex to fix the issue of invalid alias detected when a new line is added between `order` and `by` and between `group` and `by`.

I think regex might not the perfect tool here, do you think it might help to consider ANTLR for example in future?

mhewedy:
To be honest, I am not 100% comfortable with the [commit](https://github.com/spring-projects/spring-data-jpa/pull/386/commits/5f7288e0cb508d80e082117c93fd687b86e0d031) I made regarding `distinct\n`. however, all the test cases succeed. 
if you are uncomfortable as well, I can remove the commit that relates to `distinct\n`.

schauder:
Regarding REGEX alternatives: There actually is a PR open for changing this or at least large parts of the parsing to parser combinators. 
Which I guess I need to update after various issues that got fixed in the last days.

schauder:
That's squashed, polished and back ported. 

Thanks.

PR #387, opened 2019-06-24
DATAJPA-1562 - Made projections work with late Hibernate 4 versions.  

schauder:
For Projections we check the Hibernate version to make sure it supports Tuples.
This was only done based on Hibernate 5 versions although some Hibernate 4 versions do so as well.
This is now taken in consideration for the version check.

The actual versions used are determined by trial and error because no clear documentation when this was introduced to Hibernate could be found.

**This is only needed on the 1.11.x branch because the other versions only support later Hibernate versions anyway**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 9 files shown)


mp911de:

PR #388, opened 2019-07-01
DATAJPA-1544 - Delete totals.size() == 1 condition in count method

JIANGc:
The totals.get(0) cannot always get the correct count result no matter
whether there is a group by clause or not. There is also
misunderstanding when we use getTotalElements method for paging.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
@@ -15,10 +15,15 @@
  */
 package org.springframework.data.jpa.repository.query;
 
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
+import static java.util.regex.Pattern.compile;
+
 import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.persistence.EntityManager;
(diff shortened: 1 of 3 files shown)


schauder:
This change causes a couple of test failures in `UserRepositoryFinderTests` `UserRepositoryTests` and tests derived from that.

@JIANGc can you take a look if this can be saved?

JIANGc:
Thanks @schauder . I made some updates and fixed unit tests.

src/main/java/org/springframework/data/jpa/repository/query/JpaQueryExecution.java
```diff
@@ -204,13 +210,33 @@ protected Object doExecute(final AbstractJpaQuery repositoryQuery, final Object[
 		}
 
 		private long count(AbstractJpaQuery repositoryQuery, Object[] values) {
-
+			StringBuilder builder = new StringBuilder();
```
schauder: To be honest I don't really grasp what is going on here.

I guess it starts with the previously existing version of the source code, which is bogus AFAIK since a count query must always return exactly one value.
Therefore this PR should be done on top of #400.

Any parsing of queries should happen where all the other parsing happens inside `AstractJpaQuery` derivatives.

And finally and possibly most crucial: The problem is that the generation of the count query is broken so that is the location that should get fixed.

gregturn:
As stated, this needs to be redone, on top of #400. I think starting from there with a fresh PR, will do much better.

PR #390, opened 2019-07-23
DATAJPA-1579 - Support multiple out parameters with nullable output data ad hoc stored procedure

JyotirmoyVS:
The Pull request corresponds to the JIRA ticket DATAJPA-1579.

In the initial code, there had been a bug that if a stored procedure didn't set any value in the output variable(s), while mapping the output to JAVAs map, there had been a null pointer exception.

The issue has been fixed by using Map<String, Optional<Object>> instead of Map<String, Object>.

Some of the test cases were modified to incorporate these implementation changes. And one new test case has been added.

Please consider the changes.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/StoredProcedureJpaQuery.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/StoredProcedureJpaQuery.java
@@ -28,6 +28,7 @@ import javax.persistence.ParameterMode;
 import javax.persistence.StoredProcedureQuery;
 import javax.persistence.TypedQuery;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@JyotirmoyVS Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=390)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=390) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@JyotirmoyVS Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=390)!

schauder:
Thanks for reporting the issue and proposing a fix. 

Since `HashMap` does support `null` just fine and the `NullPointerException` came from a known bug in the JDK and also `Stream` and `Optional` tend to cause performance problems anyway I decided to solve the issue based on your test with an iterative approach.

See: https://stackoverflow.com/questions/24630963/java-8-nullpointerexception-in-collectors-tomap

PR #392, opened 2019-10-14
How to persist/commit entity to database immediately

antusen:
I have a need to persist(insert) a entity to database immediately when the save (or saveAndFlush) code is called.
However although the entity is created in the context it is not persisted in the database immediately.
We are using Spring Boot.
```java
public interface MessageRepository extends JpaRepository<MessageEntity, Long> {
}
````
In the Service class
```java
@Service
public class TestService {

@Autowired
    private MessageRepository messageRepository;

@Transactional
        public MessageEntity saveMessage(MessageEntity entity) throws Exception {
            entity = messageRepository.saveAndFlush(entity);
            return entity;
        }
}
```
Though the entity is created it is not persisted/committed to the database immediately.


pivotal-issuemaster:
@antusen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=392)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-jpa&pullRequestId=392) to manually synchronize the status of this Pull Request.

schauder:
Pull requests are for suggesting specific changes to a code base.

This looks more like a general usage or support question, which should go to Stackoverflow.

It also seems more about general JPA usage than Spring Data JPA

PR #396, opened 2019-11-05
DATAJPA-1619 - in/notIn predicate now accept Iterable arguments.

schauder:
https://jira.spring.io/browse/DATAJPA-1619


mp911de:

PR #400, opened 2019-11-19
DATAJPA-1635 - Removed superfluous logic from evaluation of count query.

schauder:
The logic existed only for handling a scenario that was only created by bogus unit tests.
A count query must never return anything but a single result.

https://jira.spring.io/browse/DATAJPA-1635


schauder:
Turns out the  unit test wasn't that bogus after all. The generated count query behaves as the unit test describes if the original query contains a `group by`.

PR #403, opened 2019-12-26
DATAJPA-1649 Ensure repository declares correct ID type.

stsypanov:
https://jira.spring.io/browse/DATAJPA-1649

It turned out that some of repositories in test module are declared with mismatching ID type, e. g. `UserRepository` and application context starts up correctly and repository works unless one calls `CrudRepository.findById()` method.


schauder:
Thanks, that is merged as #9d5c13e0185a0

PR #405, opened 2019-12-28
DATAJPA-1652: Using || (pipes) along with named parameters in custom queries raises an exception

reta:

Since the Lovelace release train, the usage of '||' along named parameters in custom queries ends up with `Caused by: org.hibernate.QueryException: Named parameter [..] not set` exception.  For example,

```
@Query(value = "SELECT u FROM User u WHERE u.lastname LIKE '%'||:name||'%'")
Page<User> findByNamedQueryWithLike(@Param("name") String name, Pageable page); 
```

It caused by the fact that the identifier pattern (declared in `QueryUtils`) has changed in Lovelace and above, instead of capturing 'name', it captures 'name||'.  The suggested fix is to exclude `|` from the identifier name pattern.

Fixes https://jira.spring.io/browse/DATAJPA-1652, could be also related to https://github.com/spring-projects/spring-framework/issues/22450 and https://github.com/spring-projects/spring-data-commons/pull/340. 

Thank you.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -79,6 +79,7 @@ import org.springframework.util.StringUtils;
  * @author Florian Lüdiger
  * @author Grégoire Druant
  * @author Mohammad Hewedy
+ * @author Andriy Redko
  */
 public abstract class QueryUtils {
 
@@ -90,7 +91,7 @@ public abstract class QueryUtils {
 	// Cc Control
 	// Cf Format
 	// Punct Punctuation
-	private static final String IDENTIFIER = "[._$#[\\P{Z}&&\\P{Cc}&&\\P{Cf}&&\\P{Punct}]]+";
+	private static final String IDENTIFIER = "[._$[\\P{Z}&&\\P{Cc}&&\\P{Cf}&&\\P{Punct}]]+";
 	static final String COLON_NO_DOUBLE_COLON = "(?<![:\\\\]):";
 	static final String IDENTIFIER_GROUP = String.format("(%s)", IDENTIFIER);
 
--- a/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -89,8 +89,8 @@ public abstract class QueryUtils {
 	// Z Separator
 	// Cc Control
 	// Cf Format
-	// P Punctuation
-	private static final String IDENTIFIER = "[._[\\P{Z}&&\\P{Cc}&&\\P{Cf}&&\\P{P}]]+";
+	// Punct Punctuation
+	private static final String IDENTIFIER = "[._$#[\\P{Z}&&\\P{Cc}&&\\P{Cf}&&\\P{Punct}]]+";
 	static final String COLON_NO_DOUBLE_COLON = "(?<![:\\\\]):";
 	static final String IDENTIFIER_GROUP = String.format("(%s)", IDENTIFIER);
 
--- a/src/test/java/org/springframework/data/jpa/repository/query/StringQueryUnitTests.java
+++ b/src/test/java/org/springframework/data/jpa/repository/query/StringQueryUnitTests.java
@@ -37,6 +37,7 @@ import org.springframework.data.repository.query.parser.Part.Type;
  * @author Thomas Darimont
  * @author Jens Schauder
  * @author Nils Borrmann
+ * @author Andriy Redko
  */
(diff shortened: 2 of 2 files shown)


src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -90,7 +90,7 @@
 	// Cc Control
 	// Cf Format
 	// P Punctuation
-	private static final String IDENTIFIER = "[._[\\P{Z}&&\\P{Cc}&&\\P{Cf}&&\\P{P}]]+";
```
schauder: This adds a the `|` to the list of separators for an identifiers. 
But the underlying problem lies deeper: the Unicode punctuation (`\p{P}`)category doesn't contain many of the values in the Posix punctuation category (`\{Punct}`). For example `<` and `>` aren't in that list either.

I quickly tried `\{Punct}` but breaks some other test.

In any case we need a solution addressing the general problem.

src/test/java/org/springframework/data/jpa/repository/UserCustomRepositoryFinderTests.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Integration test for executing custom finders, thus testing various query lookup strategies.
+ *
+ * @see QueryLookupStrategy
+ * @author Andriy Redko
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations = "classpath:config/namespace-customfactory-context.xml")
+@Transactional
+public class UserCustomRepositoryFinderTests {
```
schauder: I don't think we need an Integration test for this.
Adding tests in `StringQueryUnitTests` should be sufficient.

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
reta: Thanks for the feedback @schauder , I have refactored the implementation a bit to address the issue at large, wondering what do you think, all tests are passing. Thank you!

src/test/java/org/springframework/data/jpa/repository/UserCustomRepositoryFinderTests.java
reta: :+1: thanks!

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
schauder: While this fixes the problem at hand while not breaking any tests it just makes it obvious that something is really wrong here: an Identifier shouldn't contain a `#` and also probably not a `$`.

It seems `IDENTIFIER` isn't used for matching identifier, but something like identifier-or-spel 
Therefore it

a) should be named as such
b) should be constructed as such, i.e. out of two regex that matches identifiers and spels respectively.
c) all the usages should be verified that they actually should use the more general regex vs the one that actually matches identifiers.

src/test/java/org/springframework/data/jpa/repository/query/StringQueryUnitTests.java
```diff
@@ -540,6 +540,34 @@ public void isNotDefaultProjection() {
...
+		softly.assertThat(query.getParameterBindings()).hasSize(1);
+		softly.assertThat(query.getParameterBindings().get(0).getName()).isEqualTo("name");
+
+		softly.assertAll();
+	}
+
+	@Test // DATAJPA-1652
+	public void usingGreaterThanWithNamedParameter() {
+
+		String queryString = "SELECT u FROM User u WHERE :age<u.age";
```
schauder: The name says "greaterThan" but the operator used is "less than"

reta: :+1: thanks (yeah, traded `>` for `<`, expression looked a bit more comprehensible to me)

src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
reta: So the `$` seems to be a valid symbol within identifier name, there is even a dedicated test for it:
```
@Test // DATAJPA-1200
public void testHasNamedParameter() {
    ...
    checkHasNamedParameter(":$something", true, "dollar");
    ...
}
```

The `#` indeed is the SpelEx delimiter, in this context it turned out to be correct but redundant: it is explicitly included  within `NAMED_PARAMETER` pattern. So I removed it from the more strict `IDENTIFIER` match.

schauder:
Thanks, this is merged and backported.

reta:
Thanks a lot, @schauder !

PR #411, opened 2020-02-01
Adding Github Continuous Integration

GabrielBB:
Adding Github Actions Maven Workflow. This will make it easier for the owners and contributors to quickly check if pull requests are passing the maven tests. Its' free. Why not

You can see that the CI works here: https://github.com/GabrielBB/spring-data-jpa/runs/421046996

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,18 @@
+#author Gabriel Basilio
+name: Java CI
+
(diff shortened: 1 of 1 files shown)


odrotbohm:
Thanks for your suggestion. We already maintain a sophisticated build pipeline and we currently don't see any value to maintain an additional one.

PR #415, opened 2020-03-19
DATAJPA-1660 - Factory methods for JpaSort.

odrotbohm:


schauder:
Thanks, that's merged.

PR #421, opened 2020-04-30
DATAJPA-1576 - JpaMetamodelEntityInformation uses getters to access ids from proxies.

schauder:
Note that it still uses field access for non proxy entities.


mp911de:

PR #423, opened 2020-05-13
Add license information

steven-sheehy:
Add Apache 2.0 LICENSE file to project root. Also adds project license and other metadata to the pom.xml.

License file is necessary to prove the transitive dependencies in our application are the correct license when submitting our application to GCP Marketplace. Also this is a well known file that Github knows how to parse and used to show the license name in the top right.

- [ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAJPA).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        https://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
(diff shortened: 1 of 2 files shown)


mp911de:
The project-specific license is available from https://github.com/spring-projects/spring-data-jpa/blob/master/src/main/resources/license.txt. The Pom license is inherited from the parent pom. Care to elaborate on the motivation of this pull request?

steven-sheehy:
Ah, I didn't see that license.txt as it is a non-standard location. I've been filling out Google's OSS worksheet with 200+ dependencies and almost all of them have a LICENSE.txt/LICENSE/COPYING/etc. file at the root. Having it at the root also allows GitHub to parse the license. But the location in src/main/resources is sufficient for my use case as I just need to provide a link to the specific license somewhere on the project site. We can close this. Thanks

PR #424, opened 2020-06-25
DATAJPA-1800 - Wording changes

Buzzardo:
Replace potentially insensitive language with more neutral language.

Note that problematic words in the code have
to remain in the docs until the code changes.

Issue: DATAJPA-1800.


schauder:
That's merged. Thanks Jay.

PR #425, opened 2020-06-28
DATAJPA-1750 - Fixes order clause creation if function with parameter…

simonparadies:

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -81,6 +81,7 @@ import org.springframework.util.StringUtils;
  * @author Mohammad Hewedy
  * @author Andriy Redko
  * @author Peter Großmann
+ * @author Simon Paradies
  */
 public abstract class QueryUtils {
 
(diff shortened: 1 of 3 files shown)


schauder:
Thanks for looking into this. The problem, not with your PR, but with the current situation is, that the current approach of parsing the queries is not maintainable and need some serious work. 
This rework should solve the underlying problem. 

I'd like to keep the PR open though, in order to eventually merge at least the tests, in order to verify everything works as intended.

simonparadies:
`I'd like to keep the PR open though, in order to eventually merge at least the tests, in order to verify everything works as intended.`
Since this PR solves the issue at hand, why not merge it as-is now?
`The problem, not with your PR, but with the current situation is, that the current approach of parsing the queries is not maintainable and need some serious work. This rework should solve the underlying problem.`
Imo that is a different and independent problem which should not delay this fix. I'd like to get away from ugly and imperformant work-arounds asap. I am afraid that a major rework will take its time. Merging the PR now does not impede a rework in any way.

gregturn:
Backported to `2.7.x` and `2.6.x`.

Thanks @simonparadies!

PR #426, opened 2020-07-03
DATAJPA-872 - Allow for duplicate hints.

schauder:
Some JPA implementations allow to have multiple hints for the same key/name.
The former implementation did not support that since it stored hints as Maps.

With this change they are now stored as lists and duplicates are passed on to the JPA implementation.

How the implementation deals with such a situation depends on that implementation.

Issue: DATAJPA-872

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
@@ -41,6 +41,7 @@ import org.springframework.data.jpa.repository.query.JpaQueryExecution.Procedure
 import org.springframework.data.jpa.repository.query.JpaQueryExecution.SingleEntityExecution;
 import org.springframework.data.jpa.repository.query.JpaQueryExecution.SlicedExecution;
 import org.springframework.data.jpa.repository.query.JpaQueryExecution.StreamExecution;
+import org.springframework.data.jpa.repository.support.QueryHintValue;
 import org.springframework.data.jpa.util.JpaMetamodel;
 import org.springframework.data.repository.query.RepositoryQuery;
 import org.springframework.data.repository.query.ResultProcessor;
@@ -239,11 +240,11 @@ public abstract class AbstractJpaQuery implements RepositoryQuery {
 		JpaEntityGraph entityGraph = method.getEntityGraph();
 
 		if (entityGraph != null) {
-			Map<String, Object> hints = Jpa21Utils.tryGetFetchGraphHints(em, method.getEntityGraph(),
+			List<QueryHintValue> hints = Jpa21Utils.tryGetFetchGraphHints(em, method.getEntityGraph(),
 					getQueryMethod().getEntityInformation().getJavaType());
 
-			for (Map.Entry<String, Object> hint : hints.entrySet()) {
-				query.setHint(hint.getKey(), hint.getValue());
+			for (QueryHintValue hint : hints) {
+				query.setHint(hint.name, hint.value);
 			}
 		}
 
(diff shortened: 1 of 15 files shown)


src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
```diff
@@ -239,11 +240,11 @@ private Query applyEntityGraphConfiguration(Query query, JpaQueryMethod method)
 		JpaEntityGraph entityGraph = method.getEntityGraph();
 
 		if (entityGraph != null) {
-			Map<String, Object> hints = Jpa21Utils.tryGetFetchGraphHints(em, method.getEntityGraph(),
+			List<QueryHintValue> hints = Jpa21Utils.tryGetFetchGraphHints(em, method.getEntityGraph(),
```
mp911de, line 243: Let's see whether we can make this method returning `QueryHints` directly.

src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
```diff
@@ -239,11 +240,11 @@ private Query applyEntityGraphConfiguration(Query query, JpaQueryMethod method)
...
 					getQueryMethod().getEntityInformation().getJavaType());
 
-			for (Map.Entry<String, Object> hint : hints.entrySet()) {
-				query.setHint(hint.getKey(), hint.getValue());
+			for (QueryHintValue hint : hints) {
```
mp911de, line 246: Instead of iterating over  `List`/`Map`, let's provide `QueryHints.forEach(BiConsumer<String, Object>)` so we can pass in a method reference that is called for each tuple and potentially for the same key multiple times.

src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadata.java
```diff
@@ -48,20 +51,56 @@
 	 * Returns all query hints to be applied to queries executed for the CRUD method.
 	 *
 	 * @return
+	 * @deprecated use {@link #getQueryHintList()} instead.
 	 */
+	@Deprecated
 	Map<String, Object> getQueryHints();
```
mp911de, line 57: It should be fine to return `QueryHints` as this type is store-internal API that shouldn't be called externally.

src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadataPostProcessor.java
```diff
@@ -226,25 +227,25 @@ private static LockModeType findLockModeType(Method method) {
 			return annotation == null ? null : (LockModeType) AnnotationUtils.getValue(annotation);
 		}
 
-		private static Map<String, Object> findQueryHints(Method method, Predicate<QueryHints> annotationFilter) {
+		private static List<QueryHintValue> findQueryHints(Method method, Predicate<QueryHints> annotationFilter) {
```
mp911de, line 230: Let's collect hints in a `MultiValueMap`. That way, we can avoid the additional type.

src/main/java/org/springframework/data/jpa/repository/support/QueryHintValue.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * Value object carrying a query hint consisting of a name/key and a value.
+ * 
+ * @author Jens Schauder
+ * @since 2.4
+ */
+public class QueryHintValue {
```
mp911de, line 28: Let's remove this one in favor of a slightly different implementation.

src/main/java/org/springframework/data/jpa/repository/support/QueryHints.java
```diff
@@ -17,21 +17,20 @@
...
+ * QueryHints provides access to query hints defined via {@link CrudMethodMetadata#getQueryHintList()} by default
+ * excluding JPA {@link javax.persistence.EntityGraph}.
  *
  * @author Christoph Strobl
  * @author Oliver Gierke
  * @author Jens Schauder
  * @since 2.0
  */
-interface QueryHints extends Iterable<Entry<String, Object>> {
+interface QueryHints extends Iterable<QueryHintValue> {
```
mp911de, line 33: We could either remove `Iterable` entirely or use `Entry<String, List<Object>>`. Since the type is internal anyway, we're free for a bit more aggressive cleanup.

mp911de:

martinth:
Thank you @schauder for fixing the issue :) This unblocks a lot of things for me!

PR #431, opened 2020-10-15
DATAJPA-1798 - Editing pass.

Buzzardo:
Edited new content (and some old content that may have been missed).

Issue: DATAJPA-1798.


schauder:
Thanks, that's merged.

PR #432, opened 2020-10-21
DATAJPA-1803 - Fixes JavaDoc in EnableJpaAuditing from joda.time to java.time.

rd-malte-mauelshagen:


schauder:
Thanks for noting and fixing this.
This is merged.

PR #434, opened 2020-10-30
DATAJPA-1813 - Removes dependency to joda.time

Treppenhouse:
The last remaining usages in a test are replaced with a java.time equivalent.


mp911de (approved):

schauder:
Thanks, that's merged.

PR #435, opened 2020-11-13
DATAJPA-1818 - Implements CrudRepository.delete(Iterable<ID> ids).

schauder:
Merge right after DATACMNS-800

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
+++ b/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
@@ -234,7 +234,7 @@ public class SimpleJpaRepository<T, ID> implements JpaRepositoryImplementation<T
 	 */
 	@Transactional
 	@Override
-	public void deleteInBatch(Iterable<T> entities) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
```diff
@@ -218,6 +218,16 @@ public void deleteAll(Iterable<? extends T> entities) {
 		}
 	}
 
+	@Override
+	public void deleteAllById(Iterable<? extends ID> ids) {
```
mp911de, line 222: Should we also add a `deleteAllByIdInBatch` variant?

schauder: I added `deleteAllByIdInBatch`.

It does not work with EclipseLink since their handling of IN queries seems still to be broken.

Renamed `deleteInBatch` to `deleteAllInBatch` to match the naming conventions of `CrudRepository` and the new method. The old variant is deprecated and delegates to the new one.

mp911de (approved):

mp911de:

PR #438, opened 2020-12-08
DATAJPA-1827 - Consider wrapper types for Modifying JPA Query Execution

mp911de:
We now consider wrapper types (nullable types, Vavr/Javaslang/Futures) as potential wrappers and inspect the component type of each wrapper to determine the actual method return type.

---

Related ticket: DATAJPA-1827


schauder:
That's merged.

PR #2148, opened 2021-02-03
DATAJPA-1574 - Add support for saveAllAndFlush.

### LINKED ISSUE #1883: saveAndFlush could accept iterable as well [DATAJPA-1574]
**[Vladimir Nemergut](https://jira.spring.io/secure/ViewProfile.jspa?name=vlado_nemo)** opened **[DATAJPA-1574](https://jira.spring.io/browse/DATAJPA-1574?redirect=false)** and commented

JpaRepository interface only has saveAndFlush method that takes a single instance.

\<S extends T> S saveAndFlush(S var1)

Would be handy if an overloaded method exists that takes the iterable, similar to save method, e.g.:

\<S extends T> List\<S> saveAndFlush(Iterable\<S> var1)

\<S extends T> List\<S> save(Iterable\<S> var1);

---
No further details from [DATAJPA-1574](https://jira.spring.io/browse/DATAJPA-1574?redirect=false)

SKrabbenborg:
Closes #1883

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/JpaRepository.java
+++ b/src/main/java/org/springframework/data/jpa/repository/JpaRepository.java
@@ -31,6 +31,7 @@ import org.springframework.data.repository.query.QueryByExampleExecutor;
  * @author Oliver Gierke
  * @author Christoph Strobl
  * @author Mark Paluch
+ * @author Sander Krabbenborg
  */
 @NoRepositoryBean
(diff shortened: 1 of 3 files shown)


SKrabbenborg:
Would it be advisable to add a statement in the javadoc that saveAll() and flush() are used, thus overriding those methods will also change the functionality of this method?

mp911de:
We typically don't document these things. It gives us a certain freedom to change the actual implementation. You could add `@see #saveall` and `@see #flush` references.

schauder:
Thanks for your work. That's polished and merged.

PR #2207, opened 2021-04-26
Improve wording on transactional methods inherited from `SimpleJpaRepository` and repository fragments

vananiev:
Transactionality documentation sectional may mislead about apply `@Transactional` for custom repository, which extends `JpaRepository` ([proof](https://stackoverflow.com/questions/39827054/spring-jpa-repository-transactionality)). 

Spring JPA JavaDoc may be clarified by this PR.

**Tests**:
I test
```java
public interface SecurityRepository extends JpaRepository<SecurityEntity, String> {
    Collection<SecurityEntity> findByIdStartingWith(String with);
}
```
method call with property
```
logging.level.org.springframework.transaction.interceptor = TRACE
```
When no `@Transactional` on class or method, transaction for custom method is not created
```
o.s.t.i.TransactionInterceptor           : No need to create transaction for [org.springframework.data.jpa.repository.support.SimpleJpaRepository.findByIdStartingWith]: This method is not transactional.
```
With annotation on class or method level transaction is created
```
o.s.t.i.TransactionInterceptor           : Getting transaction for [org.springframework.data.jpa.repository.support.SimpleJpaRepository.findByIdStartingWith]
o.s.t.i.TransactionInterceptor           : Completing transaction for [org.springframework.data.jpa.repository.support.SimpleJpaRepository.findByIdStartingWith]
```

### DIFF
--- a/src/main/asciidoc/jpa.adoc
+++ b/src/main/asciidoc/jpa.adoc
@@ -869,7 +869,7 @@ include::query-by-example.adoc[leveloffset=+1]
 [[transactions]]
 == Transactionality
 
-By default, CRUD methods on repository instances are transactional. For read operations, the transaction configuration `readOnly` flag is set to `true`.  All others are configured with a plain `@Transactional` so that default transaction configuration applies. For details, see JavaDoc of link:$$https://docs.spring.io/spring-data/data-jpa/docs/current/api/index.html?org/springframework/data/jpa/repository/support/SimpleJpaRepository.html$$[`SimpleJpaRepository`]. If you need to tweak transaction configuration for one of the methods declared in a repository, redeclare the method in your repository interface, as follows:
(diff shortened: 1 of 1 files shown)


mp911de:
Actually, methods on `JpaRepository` and `JpaSpecificationExecutor` are transactional as `SimpleJpaRepository` is marked as `@Transactional(readOnly = true)`.

vananiev:
Yes. You give measlead answer as documentation. That's a problem. Does your answer follow that `SecurityRepository .findByIdStartingWith()` is tarnsactional? Test shows that no.

mp911de:
> Does your answer follow that SecurityRepository .findByIdStartingWith() is tarnsactional? Test shows that no.

`findByIdStartingWith(…)` is not declared on `JpaRepository` or `JpaSpecificationExecutor`, it is declared on `SecurityRepository` and likely it isn't annotated with `@Transactional`.

Therefore your observation is correct. Note that transactional attributes defined on `SimpleJpaRepository` (which is the primary implementation of `JpaRepository` and `JpaSpecificationExecutor`) do not affect query methods you declare on your repository unless you override a method of `JpaRepository` or `JpaSpecificationExecutor`.

vananiev:
Yes, but doc says about [all] CRUD methods of repository

> By default, CRUD methods on repository instances are transactional

mp911de:
> This part may be clarified by this PR.

This PR addresses only a subset, so my intent is rather to mention both interfaces to document the exact behavior to expect.

vananiev:
Ok, will you make it yourself or my help required?

mp911de:
I can take it if you don't mind.

vananiev:
Ok, but I wanto to check if explanation will be clean.

Issue #2226, opened 2021-06-07
JPA Repository unnecessarily requires transactions


l-dobrev:
TLDR; The `SimpleJpaRepository` base is marked `@Transactional` which requests a transaction for every operation, even when there is no need for that.

Applying `@Transactional(propagation = Propagation.SUPPORTS, readOnly = true)` to a service method is a way to request a read-only database operation without creating a dedicated transaction.
The use-case in mind is using PostgreSQL master-slave replication with a PgPool infront of the cluster. The pool will always create a connection to the master server if a transaction is requested (required for isolation and locking to work). Whenever a read-only process is done, not creating a wrapping transaction allows the queries to be forwarded to slave servers thus reducing the stress on the master server.

The current blank `@Transactional` annotation on the `SimpleJpaRepository` breaks this in a way that may be hard to replicate. Transactional services (Supports+Read-Only) that load entities via the JPA Repositories will have their JPA Context paused, a new nested JPA Context is created, a new Transaction is created just for the JPA method, the entity/entities are loaded within the nested context and returned. The nested context and transaction are closed, and the entities end up being used outside their native JPA Context, which ends up crashing whenever the entity's relations (lazy ones) are requested.
These exceptions are very unexpected as normally a "Session is closed" exception is thrown while an active session is available.

The reason for this is that the blank `@Transactional` annotation has more specificity than a `@Transactional(propagation = Propagation.SUPPORTS, readOnly = true)` which is on the outside.

The side effects range from unexpected "Session is closed" exceptions to situations where multiple (sometimes hundreds) of transactions are repeatedly created and committed for a single service call.

l-dobrev:
Correction: the `SimpleJpaRepository` is marked `@Transactional(readOnly = true)` no a blank `@Transactional`.

schauder:
I understand your concern regarding PostgreSQLs master-slave setup. But that does not warrant a general change in behaviour I think.

I do not understand the argument about lazy loading failing when a transaction is required, because it should fail if non is present in the first place. From the JPA spec: 

> The find method (provided it is invoked without a lock or invoked with LockModeType.NONE) and the getReference method are not required to be invoked within a transaction. If an entity man- ager with transaction-scoped persistence context is in use, the resulting entities will be detached; if an entity manager with an extended persistence context is used, they will be managed.

schauder:
I understand that, but you example should also crash without the `@Transactional` annotation on `SimpleJpaRepository`.

[...]

PR #2227, opened 2021-06-07
Fix jpa supports transaction demarcation

l-dobrev:

This is a suggested resolution for the #2226 issue


mp911de:
As per https://github.com/spring-projects/spring-data-jpa/issues/2226#issuecomment-862393406, we're closing this pull request since the propagation mode isn't broken.

PR #2229, opened 2021-06-08
Escape JDBC styled parameters.

### LINKED ISSUE #2228: ExpressionBasedStringQuery does not quote all parameter expressions.
I haven't tested this yet, but I am sure that `ExpressionBasedStringQuery.java` won't correctly quote parameters in the form `:#{#something}` which are allowed in addition to `?#{#something}`. See Spring Data commons: https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/repository/query/SpelQueryContext.java#L46

michael-simons:
This fixes #2228  and is necessary to work with SpEl parameters that can be given with both `:` and `?`.


schauder:
That's merged. Thanks Michael.

PR #2231, opened 2021-06-09
Fix build failures with Java 16.

schauder:
A few test that used reflection to access the actual SQL executed by a Query object failed with `java.lang.reflect.InaccessibleObjectException: Unable to make field protected java.lang.reflect.InvocationHandler java.lang.reflect.Proxy.h accessible: module java.base does not "opens java.lang.reflect" to unnamed module 1b9e1916`.
This change replaces the reflection used by a proper call to `unwrap`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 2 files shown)


mp911de:
Feel free to merge this right away as these are test changes only.

schauder:
Merged.

PR #2245, opened 2021-06-22
Drop serialVersionUID from AbstractPersistable.

quaff:
Since serialVersionUID already exists

### DIFF
--- a/src/main/java/org/springframework/data/jpa/domain/AbstractPersistable.java
+++ b/src/main/java/org/springframework/data/jpa/domain/AbstractPersistable.java
@@ -36,7 +36,7 @@ import org.springframework.lang.Nullable;
  * @param <PK> the type of the identifier.
  */
 @MappedSuperclass
-public abstract class AbstractPersistable<PK extends Serializable> implements Persistable<PK> {
+public abstract class AbstractPersistable<PK extends Serializable> implements Persistable<PK>, Serializable {
 
 	private static final long serialVersionUID = -5554308939380869754L;
 


schauder:
Could you explain what problem gets solved by making it `Serializable`? Apart from the obvious one of an otherwise out of place `serialVersionUID`?

quaff:
> Could you explain what problem gets solved by making it `Serializable`? Apart from the obvious one of an otherwise out of place `serialVersionUID`?

`serialVersionUID` is present here, I guess someone missed `Serializable` here, `Serializable` has no sin, It's reasonable to mark model as serializable.
If project maintainer object this PR for security concern, please close it and remove unused `serialVersionUID`.

mp911de:
Implementing `Serializable` would suddenly make all subclasses of `AbstractPersistable` serializable which isn't an intended effect. Having `serialVersionUID` helps with the object version and aids subclasses that want to rely on `serialVersionUID` and explicitly make their subclasses serializable. That being said, we should keep things as they are right now.

quaff:
> Implementing `Serializable` would suddenly make all subclasses of `AbstractPersistable` serializable which isn't an intended effect. Having `serialVersionUID` helps with the object version and aids subclasses that want to rely on `serialVersionUID` and explicitly make their subclasses serializable. That being said, we should keep things as they are right now.

It seems `serialVersionUID` doesn't inherited by subclasses.
https://stackoverflow.com/questions/23465613/is-serialversionuid-inherited-by-subclasses-if-i-have-default-serialversionuid

gregturn (commented):
Based upon @mp911de 's comments, we're actually going to drop serialVersionUID, and verify that doesn't break anything.

gregturn:
Resolved via 2e30a353.

gregturn:
Backport with d3a12428b75024fd9dd30438bbb89a9fc0d24724.

gregturn:
Backported to 2.4.x with 855bff61e144c84ec58d60573a78054bcca6b37f.

Issue #2282, opened 2021-08-20
Eliminate unnecessary "where true=1" for QBE


quaff:
`JpaRepository.findAll(Example)` will generate unnecessary `where true=1` if matched predicates is empty, It's possible to eliminate it.
https://github.com/spring-projects/spring-data-jpa/blob/acf01b6c7140a56cd1c3caa67409b4d4f922937d/src/main/java/org/springframework/data/jpa/convert/QueryByExamplePredicateBuilder.java#L106-L108

quaff:
I create PR #2283
And I have 2 other proposals:
1. check the predicate if it is equal to `cb.isTrue(cb.literal(true))` then return `null`, drawback it is not possible by JPA api, it depends on vender implementation.
https://github.com/spring-projects/spring-data-jpa/blob/acf01b6c7140a56cd1c3caa67409b4d4f922937d/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java#L939-L941
2. return `null` instead of `cb.isTrue(cb.literal(true))`, drawback is it breaks contract
https://github.com/spring-projects/spring-data-jpa/blob/acf01b6c7140a56cd1c3caa67409b4d4f922937d/src/main/java/org/springframework/data/jpa/convert/QueryByExamplePredicateBuilder.java#L106-L108

mp911de:
Thanks for reaching out. Care to elaborate what issue `true=1`is causing for you and why it would make sense to remove it?

quaff:
> Thanks for reaching out. Care to elaborate what issue `true=1`is causing for you and why it would make sense to remove it?

```java
	@GetMapping("/customers")
	public List<Customer> list(Customer example) {
		return customerRepository.findAll(Example.of(example));
	}
```
`example` object will be empty if no parameters present, and spring-data-jpa will append a meaningless `where ?=1` and passing `true`

mp911de:
I agree with aesthetics here, but does `true=1` prevent any functionality or does it break something?

quaff:
> I agree with aesthetics here, but does `true=1` prevent any functionality or does it break something?

It doesn't prevent any functionality, but I think it could harm a little bit performance (not verified).

mp911de:
Typically, databases optimize these conditions away.

gregturn:
@schauder I have drafted a modification that essentially swaps `where true=1` with a `null`. The one test case this broke I fixed. But I'm not quite sure this adequately verifies the change works.

[...]

PR #2283, opened 2021-08-20
Eliminate unnecessary "where true=1" for QBE

quaff:
fix gh-2282


quaff:
I fixed it by overloading `QueryByExamplePredicateBuilder::getPredicate`, please review. @mp911de

mp911de:
Introducing a new flag to indicate the semantics of return values creates complexity in the wrong place. We can make the `getPredicate` nullable with the 3.0 release.

schauder:
If we do, we have to clarify the meaning in the context of both conjunction and disjunction including the case where the null predicate is the only element in the con/dis-junction.

PR #2297, opened 2021-09-13
Clarify keyword not supported by JPA

quaff:
People may be confused that keyword is listed in reference but illegal here

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryCreator.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryCreator.java
@@ -322,7 +322,7 @@ public class JpaQueryCreator extends AbstractQueryCreator<CriteriaQuery<? extend
 					return type.equals(IS_NOT_EMPTY) ? builder.isNotEmpty(collectionPath) : builder.isEmpty(collectionPath);
 
 				default:
-					throw new IllegalArgumentException("Unsupported keyword " + type);
+					throw new IllegalArgumentException("Unsupported keyword " + type.name() + " by JPA");
 			}
 		}
 


schauder:
Thanks for your contribution. I don't really see what this clarifies. The exception is already thrown by Spring Data JPA which is clearly documented by the package name. 

Also the new error message is technically wrong since the problem is not JPA but Spring Data JPA not supporting the keyword.

quaff:
> Thanks for your contribution. I don't really see what this clarifies. The exception is already thrown by Spring Data JPA which is clearly documented by the package name.
> 
> Also the new error message is technically wrong since the problem is not JPA but Spring Data JPA not supporting the keyword.

I see `REGEX` in `spring-data-jpa` reference https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#appendix.query.method.subject
but exception thrown when I try to use it:
```java
Caused by: java.lang.IllegalArgumentException: Unsupported keyword REGEX (1): [MatchesRegex, Matches, Regex]
	at org.springframework.data.jpa.repository.query.JpaQueryCreator$PredicateBuilder.build(JpaQueryCreator.java:325)
	at org.springframework.data.jpa.repository.query.JpaQueryCreator.toPredicate(JpaQueryCreator.java:211)
	at org.springframework.data.jpa.repository.query.JpaQueryCreator.create(JpaQueryCreator.java:124)
	at org.springframework.data.jpa.repository.query.JpaQueryCreator.create(JpaQueryCreator.java:59)
	at org.springframework.data.repository.query.parser.AbstractQueryCreator.createCriteria(AbstractQueryCreator.java:119)
	at org.springframework.data.repository.query.parser.AbstractQueryCreator.createQuery(AbstractQueryCreator.java:95)
	at org.springframework.data.repository.query.parser.AbstractQueryCreator.createQuery(AbstractQueryCreator.java:81)
	at org.springframework.data.jpa.repository.query.PartTreeJpaQuery$QueryPreparer.<init>(PartTreeJpaQuery.java:217)
	at org.springframework.data.jpa.repository.query.PartTreeJpaQuery$CountQueryPreparer.<init>(PartTreeJpaQuery.java:348)
	at org.springframework.data.jpa.repository.query.PartTreeJpaQuery.<init>(PartTreeJpaQuery.java:91)
	... 111 more
```
`REGEX` is not supported by spring-data-jpa because JPA doesn't support it.

schauder:
> REGEX is not supported by spring-data-jpa because JPA doesn't support it.

While I assume this is correct, the error message should not speculate on the reasons why something was not implemented.
Instead it should explain why an exception was thrown. 
And the reason for that is that Spring Data JPA does not support the REGEX keyword. 
Or to put it the other way round. If the next version of JPA does support regular expressions it still doesn't make Spring Data JPA support the `REGEX` keyword in derived queries.

The part of the documentation you are referring to applies to Spring Data in general and it explicitly asks to "_Consult the store-specific documentation for the exact list of supported keywords, because some keywords listed here might not be supported in a particular store._" 

The list of keywords supported by Spring Data JPA is given in Table 3 here: https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods.query-creation

PR #2324, opened 2021-10-04
changes based on the results of static code analysis by Findbugs

dk2k:
As for removed Assert - its condition is always true, can be checked in Intellij Idea as well

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/config/JpaRepositoryConfigExtension.java
+++ b/src/main/java/org/springframework/data/jpa/repository/config/JpaRepositoryConfigExtension.java
@@ -16,6 +16,8 @@
 package org.springframework.data.jpa.repository.config;
 
 import java.lang.annotation.Annotation;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


dk2k:
What do I do now to restart checks?

src/main/java/org/springframework/data/jpa/repository/config/JpaRepositoryConfigExtension.java
```diff
@@ -234,9 +237,17 @@ public void registerBeansForRoot(BeanDefinitionRegistry registry, RepositoryConf
 	@Override
 	protected ClassLoader getConfigurationInspectionClassLoader(ResourceLoader loader) {
 
-		ClassLoader classLoader = loader.getClassLoader();
-
-		return classLoader != null && LazyJvmAgent.isActive(loader.getClassLoader())
+        AtomicReference<ClassLoader> classLoaderRef = new AtomicReference<>();
+        AccessController.doPrivileged(new PrivilegedAction() {
```
mp911de, line 241: `AccessController` is subject to deprecation and future removal with the deprecation of Java's Security Manager.

dk2k: will revert related changes then

schauder:
Thanks, that's merged.

PR #2326, opened 2021-10-06
Implement FluentQuery support for Querydsl and Query by Example

gregturn:


gregturn:
Packages moved, visibility adjusted.

mp911de:

PR #2343, opened 2021-10-21
add protectedGetEntityManager method to JpaRepository.

bokshitsky:
Hi,
(excuse me for my english :)

I am new spring-data user.

We use it for some services in our company and it works great.

But I found what if I want to have some customization of repository method and write it's body - I need to create extra classes (Impl and Custom) and put what logic there, separately from common repository logic.
(I mean this part: https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.single-repository-behavior)

I was trying to avoid same separation using Java8 default Interface methods, and found what the only missing thing for 95% of logic is EntityManager, unavailable in current JpaRepository interface.

So I prepared this PR allowing to access EntityManager from JpaRepository.

This allows to write same custom methods directly in `YourRepositoryInterface extends JpaRepository` without additional classes and logic-split.
(see TestCase in PR for example)

I don't know if this is good idea or not, so I prepared some code (this PR) to ask what do you think about same proposal.
(Probably, we may use not directly my code, but the idea of using `interface default methods` to solve problem for the majority of cases, instead of making additional classes and splitting repository logic)

I will be grateful for any comments

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/JpaRepository.java
+++ b/src/main/java/org/springframework/data/jpa/repository/JpaRepository.java
@@ -33,6 +33,7 @@ import org.springframework.data.repository.query.QueryByExampleExecutor;
  * @author Mark Paluch
  * @author Sander Krabbenborg
  * @author Jesse Wouters
+ * @author Evgeny Bokshitsky
  */
 @NoRepositoryBean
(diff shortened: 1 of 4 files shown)


mp911de:
Thank you for your suggestion. We do not want to expose the `EntityManager` through the Repository API as it is an implementation detail. If you need access to the `EntityManager`, please inject it either through constructor creation or via `@Autowired` into your custom implementation.

PR #2364, opened 2021-11-24
fixing flaky tests in UserRepositoryStoredProcedureIntegrationTests.java 

FanYuliang:
Hi, 

These two flaky tests (`entityAnnotatedCustomNamedProcedurePlus1IO2` and `entityAnnotatedCustomNamedProcedurePlus1IOoptional`) are detected by using [NonDex](https://github.com/TestingResearchIllinois/NonDex) using the following command: `mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.springframework.data.jpa.repository.UserRepositoryStoredProcedureIntegrationTests`

These two tests are flaky because `Map<String, Integer> result` is not in order, and therefore the `containsExactly` may be flaky. I updated it into `containsOnly` to make sure the assertion is order independent.

### DIFF
--- a/src/test/java/org/springframework/data/jpa/repository/UserRepositoryStoredProcedureIntegrationTests.java
+++ b/src/test/java/org/springframework/data/jpa/repository/UserRepositoryStoredProcedureIntegrationTests.java
@@ -96,7 +96,7 @@ public class UserRepositoryStoredProcedureIntegrationTests {
 
 		Map<String, Integer> result = repository.entityAnnotatedCustomNamedProcedurePlus1IO2(1);
 
-		assertThat(result).containsExactly(entry("res", 2), entry("res2", 3));
+		assertThat(result).containsOnly(entry("res", 2), entry("res2", 3));
 	}
 
 	@Test // DATAJPA-1579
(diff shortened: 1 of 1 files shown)


quaff:
`Map` implementation could be `LinkedHashMap` which is ordered, do you encounter failed tests or only detected by NonDex?

FanYuliang:
The tests works fine on my JVM, but NonDex will shuffle the order of elements in `Map` to simulate the indeterministic feature so that developers can find the potential flaky tests. Since the implementation of `Map` could also be `Hashmap`, I think it will be safer to use `containsOnly` to prevent the potential flakiness due to different implementation of `Map`.

quaff:
> The tests works fine on my JVM, but NonDex will shuffle the order of elements in `Map` to simulate the indeterministic feature so that developers can find the potential flaky tests. Since the implementation of `Map` could also be `Hashmap`, I think it will be safer to use `containsOnly` to prevent the potential flakiness due to different implementation of `Map`.

What if ordering is mandatory here?

FanYuliang:
If order is maintained (`Map` uses `LinkedHashMap` for example), then there won't no flakiness in the test. But I didn't find any order consistency in the `NamedStoredProcedureQuery` : as described in the [docs](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.stored-procedures), 

> If there are multiple out parameters specified in a @NamedStoredProcedureQuery annotation those can be returned as a Map with the key being the parameter name given in the @NamedStoredProcedureQuery annotation.

I didn't find in the doc saying that the `Map` uses implementation of `LinkedHashMap`, so I assume the output `Map` should be indeterministic and might be flaky.

schauder:
Thanks, that's merged.

Issue #2333, opened 2021-10-12
`QuerydslJpaPredicateExecutor.exists()` solves the existence test non-optimally


emmanuelidi:
The current implementation of `QuerydslJpaPredicateExecutor.exists(Predicate predicate)` appears to generate a `SELECT count(..) FROM ... WHERE <predicate>` query and then tests if this result is `> 0` to determine if there are any results matching the predicate.

For larger datasets, this is a bit inefficient. It seems that one could do a `SELECT * FROM ... WHERE <predicate> LIMIT 1` and test if any results are returned.

Does this sound reasonable? If so, I can create a PR to try and address.

PR #2368, opened 2021-12-01
Optimize SimpleJpaRepository.exists(Example) for better performance

quaff:
1. Set max results to 1 to avoid full scan
2. Select 1 to reduce size of ResultSet


quaff:
Inspired by https://github.com/spring-projects/spring-data-jpa/issues/2333

schauder:
Thanks for the PR. I polished and merged it.

PR #2371, opened 2021-12-03
Replace For-Loop with Stream

quaff:
It's a simple improvement.


schauder:
We try to avoid `Stream` and `Optional` on hot code paths like query execution since often they come with a performance cost.

Therefore we won't accept this PR.

PR #2372, opened 2021-12-03
Eliminate unnecessary summation

quaff:
count query will always return one row here

If this PR is accepted then #2371 should be discarded.

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
+++ b/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
@@ -880,14 +880,7 @@ public class SimpleJpaRepository<T, ID> implements JpaRepositoryImplementation<T
 
 		Assert.notNull(query, "TypedQuery must not be null!");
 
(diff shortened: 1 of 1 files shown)


schauder:
There are indeed cases where the query can return more than one value.
Therefore we can't accept this PR.

quaff:
> There are indeed cases where the query can return more than one value. Therefore we can't accept this PR.

It only count from `getCountQuery(Specification,Class)`, does `Specification` support `group by`?
https://github.com/spring-projects/spring-data-jpa/blob/bfa7be315a9ca69dcd4cbf76a539347bbec88b3f/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java#L812-L829

schauder:
Our official stance has for some time now that Specifications are only to be used for Where clauses. Unfortunately we haven't been always this clear about it. Therefore there is all kinds of code out there doing "interesting" stuff and we don't want to break it.

We have a proper fix for this kind of stuff in the pipeline, but unfortunately it takes some time.

PR #2373, opened 2021-12-03
Optimize QuerydslJpaPredicateExecutor.exists(Predicate)

quaff:
Fix GH-2333


quaff:
Similar to #2368

schauder:
Thanks, that's merged and backported.

PR #2376, opened 2021-12-04
Fix total count for paged specifications with GROUP BY

### LINKED ISSUE #1296: Paged queries with GROUP BY clauses return the incorrect total count when there is only 1 result [DATAJPA-945]
**[David Linfield](https://jira.spring.io/secure/ViewProfile.jspa?name=david.linfield@softwire.com)** opened **[DATAJPA-945](https://jira.spring.io/browse/DATAJPA-945?redirect=false)** and commented

For example, if we have a data model in which one company has many products, and we wish to get pages of results listing the company name and the count of the number of products each company has as follows:

```
@Query(value = "select new uk.co.example.models.CompanyWithProductCount(c.id, c.name, cast(count(p) as integer)) " +
    "from Company c join c.products as p " +
    "group by c.id ")
Page<CompanyWithProductCount> getCompaniesWithProductCount(Pageable pageable);
```

The query performed to count the total number of results simply performs a count query with a `GROUP BY` clause, hence effectively returning a list of `COUNT(p)` for each Company. Where there are multiple companies there seems to be some special case handling where the total returned the the `PageImpl` object is actually the number of rows returned by this query (the correct answer), however when there is **only one company** returned, the total returned to the `PageImpl` object is actually the value returned by the `COUNT` query (i.e. the number of products that company has).

This bug is often concealed, because `PageImpl` uses the number of items in the `Content` of the `Page` if the supplied total is less than the page size provided. So to recreate in the case above you there must be more than `pageSize` Products associated with the Company

---

**Affects:** 1.10.1 (Hopper SR1)

0 votes, 6 watchers

### LINKED ISSUE #1858: count method in JpaSpecificationExecutor ignores GROUP BY and returns wrong result. [DATAJPA-1573]
**[Dimitri](https://jira.spring.io/secure/ViewProfile.jspa?name=newdimitri)** opened **[DATAJPA-1573](https://jira.spring.io/browse/DATAJPA-1573?redirect=false)** and commented

I have this code:

```java
public Predicate toPredicate(Root<User> root, CriteriaQuery<?> query, CriteriaBuilder builder){ 
... 
query.groupBy(root.get("id")); 
... 
}
```

Method

```java
Repository.count(specification)
```

ignores GROUP BY and returns wrong result, while

```java
Repository.findAll(specification).size()
```

returns right result.

This issue looks similiar to this problem: https://jira.spring.io/browse/DATAJPA-656

---

**Affects:** 2.1.9 (Lovelace SR9)

**Attachments:**
- [datajpa-1573.zip](https://jira.spring.io/secure/attachment/27369/datajpa-1573.zip) (_89.10 kB_)

### LINKED ISSUE #2361: When use specification with group by pagination is broken
Hi, guys, faced the same issue. Unfortunately, this "feature" totally breaks the pagination. 
The problem is when I'm using 'groupBy' than a number of returned elements is much less than the totalElements and as a result totalPages is wrong too. 

List<Long> totals = query.getResultList();
in my case totals looks like
 col_0_0_ 
----------
        2
        2
        1
        1
(4 rows)
So, the content will have only 4 elements.

But in the end, total will be 6
for (Long element : totals) {
	total += element == null ? 0 : element;
}

_Originally posted by @pkuzmenkov in https://github.com/spring-projects/spring-data-jpa/issues/1034#issuecomment-971325551_

robin850:
Hello,

This patch tries to fix using paginated results found through a custom `Specification` that relies on a `GROUP BY` clause.

I hope the test is clear enough, but just in case, the problem is that since the `getCountQuery` method of `SimpleJpaRepository` takes the exact same query produced by the specification but injects a `COUNT` in the `SELECT`, the `COUNT` is done on each group of records so we end up counting them and not the "root" ones.

Using `DISTINCT` does the trick here. Actually I can't think of a case where using `DISTINCT` would be problematic so in my mind the patch could be to totally remove the `if/else` statement and rely on `countDistinct` anyway but maybe I'm blinded with my application and that's a bit too rash.

For the record, the `if/else` statement has been added in https://github.com/spring-projects/spring-data-jpa/commit/a74fadd2344e1358a5ce85b98ff3edf8e3e0a367.

Closes #1296.
Closes #2361.
Closes #1858.

Have a nice day.

------

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
+++ b/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
@@ -81,6 +81,7 @@ import org.springframework.util.Assert;
  * @author Jesse Wouters
  * @author Greg Turnquist
  * @author Yanming Zhou
+ * @author Robin Dupret
  */
 @Repository
 @Transactional(readOnly = true)
@@ -816,7 +817,7 @@ public class SimpleJpaRepository<T, ID> implements JpaRepositoryImplementation<T
 
 		Root<S> root = applySpecificationToCriteria(spec, domainClass, query);
 
-		if (query.isDistinct()) {
+		if (query.isDistinct() || query.getGroupList().size() > 0) {
 			query.select(builder.countDistinct(root));
 		} else {
 			query.select(builder.count(root));
--- a/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
+++ b/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
@@ -94,6 +94,7 @@ import org.springframework.transaction.annotation.Transactional;
  * @author Sander Krabbenborg
  * @author Jesse Wouters
  * @author Greg Turnquist
+ * @author Robin Dupret
  */
 @ExtendWith(SpringExtension.class)
 @ContextConfiguration("classpath:application-context.xml")
@@ -112,6 +113,7 @@ public class UserRepositoryTests {
 	private User fourthUser;
 	private Integer id;
 	private Role adminRole;
+	private Role managerRole;
 
 	@BeforeEach
 	void setUp() throws Exception {
@@ -126,6 +128,7 @@ public class UserRepositoryTests {
 		fourthUser = new User("kevin", "raymond", "no@gmail.com");
 		fourthUser.setAge(31);
 		adminRole = new Role("admin");
+		managerRole = new Role("manager");
 
 		SampleSecurityContextHolder.clear();
 	}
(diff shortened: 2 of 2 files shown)


schauder:
Sorry for being so late to reply to the original issue. 
As explained there we will not consider this PR.

ankitpec72:
@robin850 @schauder We are facing the same issue in Spring-data-jpa-2.7.13 version. Is there any solution or this issue is already fixed in some other version? The count query with group by is returning wrong total count value.

robin850:
@ankitpec72 : No, as stated in @schauder's comment, the fact that it doesn't work is "normal" because you should only manage `WHERE` clauses through `Specification`.

To be honest, we currently have a copy of this PR's `SimpleJpaRepository.java` file in our application to be able to correctly use `GROUP BY`. It's far from ideal but I don't know if there is an "elegant" API currently to achieve the same thing.

PR #2377, opened 2021-12-04
`@Query` supports query derivation from substitute.

DiegoKrupitza:
Method names for derived queries can become quite long if a property of the class has a long name, even though the logic behind the query is trivial. For such simple queries we now can provide the long and maybe ugly name as a substitute via the `@Query` annotation and have a more pleasant/short and useable method name. The `PartTreeJpaQuery` will use the substitute to derive the query.

Example Use case:

`Course.java`
```java
class Course {
   private LocalDateTime preferenceSubDeadline;
   private LocalDateTime teamSubDeadline;
   //...
}
```

`CourseRepositoryWithLongName.java`
```java
@Repository
public interface CourseRepositoryWithLongName
    extends PagingAndSortingRepository<Course, Long> {

    List<Course> findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(LocalDateTime pref, LocalDateTime team);

}
```

The super long but actually trivial method to find all courses with preference deadline before X and team deadline after Y inside `CourseRepositoryWithLongName` could be simplified by using the now introduced "substitute" way.

`CourseRepositoryWithSubstitute.java`
```java
@Repository
public interface CourseRepositoryWithSubstitute
    extends PagingAndSortingRepository<Course, Long> {

    @Query(substitute = "findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter")
    List<Course> inbetweenPrefAndTeamDeadline(LocalDateTime pref, LocalDateTime team);

}
```

As you can see the `CourseRepositoryWithSubstitute` is now cleaner and still does not required the programmer to write a custom query. 

I often encountered the case that the method name became quite long although the idea behind the method is trivial. I always thought writing a query by hand would be too much, since for most trivial cases the derived query is more convenient.

PS: I did not know if I should add myself to the list of authors since I did not know if my change is worthy enough 😄 .

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/Query.java
+++ b/src/main/java/org/springframework/data/jpa/repository/Query.java
@@ -45,7 +45,8 @@ public @interface Query {
 
 	/**
 	 * Defines a special count query that shall be used for pagination queries to lookup the total number of elements for
-	 * a page. If none is configured we will derive the count query from the original query or {@link #countProjection()} query if any.
+	 * a page. If none is configured we will derive the count query from the original query or {@link #countProjection()}
+	 * query if any.
 	 */
 	String countQuery() default "";
 
@@ -77,4 +78,11 @@ public @interface Query {
 	 * @return
 	 */
 	String countName() default "";
+
+	/**
+	 * The substitute to be used instead of the method name to derive the query
+	 * 
+	 * @return the substitue
+	 */
+	String substitute() default "";
 }
--- a/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
@@ -92,6 +93,7 @@ public class JpaQueryMethod extends QueryMethod {
 	private final Lazy<Boolean> isCollectionQuery;
 	private final Lazy<Boolean> isProcedureQuery;
 	private final Lazy<JpaEntityMetadata<?>> entityMetadata;
+	private final Lazy<String> substituteQuery;
 
 	/**
 	 * Creates a {@link JpaQueryMethod}.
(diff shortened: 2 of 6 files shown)


quaff:
I think default method is better
```java

@Repository
public interface CourseRepositoryWithSubstitute
    extends PagingAndSortingRepository<Course, Long> {

    List<Course> findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(LocalDateTime pref, LocalDateTime team);

    default List<TestEntity> inbetweenPrefAndTeamDeadline(LocalDateTime pref, LocalDateTime team) {
        return findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(pref, team);
    }
}

```

DiegoKrupitza:
> I think default method is better
> 
> ```java
> 
> 
> 
> @Repository
> 
> public interface CourseRepositoryWithSubstitute
> 
>     extends PagingAndSortingRepository<Course, Long> {
> 
> 
> 
>     List<Course> findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(LocalDateTime pref, LocalDateTime team);
> 
> 
> 
>     default List<TestEntity> inbetweenPrefAndTeamDeadline(LocalDateTime pref, LocalDateTime team) {
> 
>         return findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(pref, team);
> 
>     }
> 
> }
> 
> 
> 
> ```

Of course a default method can be used here but this also means adding "boilerplate" code that basically just blows up our repository. With the @Query we have the benefit to keep our Interfaxe small in size.

quaff:
> > I think default method is better
> > ```java
> > @Repository
> > 
> > public interface CourseRepositoryWithSubstitute
> > 
> >     extends PagingAndSortingRepository<Course, Long> {
> > 
> > 
> > 
> >     List<Course> findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(LocalDateTime pref, LocalDateTime team);
> > 
> > 
> > 
> >     default List<TestEntity> inbetweenPrefAndTeamDeadline(LocalDateTime pref, LocalDateTime team) {
> > 
> >         return findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(pref, team);
> > 
> >     }
> > 
> > }
> > ```
> 
> Of course a default method can be used here but this also means adding "boilerplate" code that basically just blows up our repository. With the @query we have the benefit to keep our Interfaxe small in size.

I think your solution will increase unnecessary complexity from spring-data-jpa perspective, and lose compile-time checking (eg. rename refactoring) from application perspective.

DiegoKrupitza:
> > > I think default method is better
> > > ```java
> > > @Repository
> > > 
> > > public interface CourseRepositoryWithSubstitute
> > > 
> > >     extends PagingAndSortingRepository<Course, Long> {
> > > 
> > > 
> > > 
> > >     List<Course> findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(LocalDateTime pref, LocalDateTime team);
> > > 
> > > 
> > > 
> > >     default List<TestEntity> inbetweenPrefAndTeamDeadline(LocalDateTime pref, LocalDateTime team) {
> > > 
> > >         return findAllByPreferenceSubDeadlineBeforeAndTeamSubDeadlineAfter(pref, team);
> > > 
> > >     }
> > > 
> > > }
> > > ```
> > 
> > 
> > Of course a default method can be used here but this also means adding "boilerplate" code that basically just blows up our repository. With the @query we have the benefit to keep our Interfaxe small in size.
> 
> I think your solution will increase unnecessary complexity from spring-data-jpa perspective, and lose compile-time checking (eg. rename refactoring) from application perspective.

I totally see your point, but it still feels quite `hacky` to me to use a default implementation 😅 . 
Just to clarify: is the idea behind substitute introducing to much complexity or the current implementation?

gregturn:
In general, when people wish to write shorthand variants of query methods, we recommend they use the `@Query`. This effectively side steps query derivation and simply lets the given method name (whatever you want) be linked to the entered query and its associated arguments.

To expand the functionality with some algorithmic solution, while module-specific, may imply a functionality that's expected in other modules as well. And generally, the Spring Data team doesn't wish to maintain such a feature.

PR #2378, opened 2021-12-06
Fix obsoleted package name in javadoc

quaff:
`com.mysema.query` renamed to `com.querydsl.core`


schauder:
Thanks, that's merged and back ported.

PR #2379, opened 2021-12-06
Destroy docker container properly

quaff:


schauder:
Thanks for the PR. 

Maybe I'm a little slow on Monday morning, but I fail to see what problem this PR is solving. Could you please elaborate?

quaff:
> Thanks for the PR.
> 
> Maybe I'm a little slow on Monday morning, but I fail to see what problem this PR is solving. Could you please elaborate?

After this commit, container which implements `AutoCloseable` is created as spring bean, it will be closed when spring context close.

schauder:
Thanks. That's merged.

Issue #2177, opened 2021-03-14
Using native Query with distinct is returning wrong total count 


OctavioCega:
Hello, It seems that Hibernate is removing the "distinct" key word from the native Query when using the count to get the total elements. 

Any Workaround for this?

Springboot 2.3.7

A fragment for the logs:

The log for the get entities
```
select distinct e.* from schema.entity e 
(more statements...)
...
```

The log for the count query generated by hibernate (as you can see, for some reason, it is removing the "distinct" keyword)
```
select count(e) from  schema.entity e 
(more statements...)
...
```

This seems like bug for me. I saw this issue from a long ago time but without any solution (a part from generating my own count query) https://github.com/spring-projects/spring-data-jpa/issues/750

ruslanys:
Looks like changes leads to the following:  https://github.com/spring-projects/spring-data-jpa/issues/2260#issuecomment-1111986155

ruslanys:
I found out that this commit has bug: https://github.com/spring-projects/spring-data-jpa/commit/3e64d9ad9b7d45fcf1231dbaf207be49cc481e7a

Especially, the line 494: `&& !variable.contains("*");`

The invocation of method

```
createCountQueryFor("SELECT * FROM User WHERE created_at > $1")
```

returns `select count(WHERE) FROM User WHERE created_at > $1`

ruslanys:
Commits and tests marked with issue #2341

DiegoKrupitza:
> I found out that this commit has bug: https://github.com/spring-projects/spring-data-jpa/commit/3e64d9ad9b7d45fcf1231dbaf207be49cc481e7a

Ok I will take a look at this. I think I have a fix for this. PR will come soon.

ruslanys:
@DiegoKrupitza Thank you!

PR #2389, opened 2021-12-18
GH-2341 - Fix missing `DISTINCT` in count query bug.

### LINKED ISSUE #2341: Count query removes `distinct` phrase if from part of JPA query contains newline character
It seems that there is newline character after the entity alias name in `FROM` part of the query before the `JOIN` the distinct is removed from the count.

Incorrect behavior:
 
```java
@Query("""
    SELECT DISTINCT entity1
    FROM Entity1 entity1 
    LEFT JOIN Entity2 entity2 ON entity1.key = entity2.key""")
```

returns
 
```
select count(entity1) FROM Entity1 entity1
LEFT JOIN Entity2 entity2 ON entity1.key = entity2.key
``` 

Correct behaviour:
```java
@Query("""
    SELECT DISTINCT entity1
    FROM Entity1 entity1 LEFT JOIN Entity2 entity2 ON entity1.key = entity2.key""")
```

returns
 
```
select count(DISTINCT entity1) FROM Entity1 entity1 LEFT JOIN Entity2 entity2 ON entity1.key = entity2.key
``` 

From testing if the `FROM` and `JOIN` are in the same line, and `WHERE` is new line then this incorrect behavior  also occurs.

For now I have solved it with `countProjection = "distinct entity1"`.

DiegoKrupitza:
The `COUNT_MATCH` did not consider line breaks after the `from` clause or the `where` clause. This lead to a no match scenario in the construction of the count query. With the fix we now consider line breaks/whitespaces after the `from` and `where` clause.

Closes #2341
Related tickets #2177

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -146,7 +146,7 @@ public abstract class QueryUtils {
 		builder.append(IDENTIFIER_GROUP);
 		builder.append("(.*)");
 
-		COUNT_MATCH = compile(builder.toString(), CASE_INSENSITIVE);
+		COUNT_MATCH = compile(builder.toString(), CASE_INSENSITIVE | DOTALL);
 
 		Map<PersistentAttributeType, Class<? extends Annotation>> persistentAttributeTypes = new HashMap<>();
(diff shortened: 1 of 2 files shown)


gregturn:
Resolved via 2545e6d74a550d748c4b59a7bf0972ffc7d547bd.

Thanks @DiegoKrupitza!

Issue #2393, opened 2021-12-31
Incorrectly generating count queries for queries starting with whitespaces.


jaoromi:
Repository method annotated `@Query`, nativeQuery is true and containing pageable parameter generate incorrectly count query for starting with whitespaces.

For example, above method derived counter query `select count(where) from user where user.age > :age `.
```
@Query("   \n"
+ "select * from user where user.age > :age"
+ "\n    ")
Page<User> findByGreaterThanAge(int age, Pageable pageable);
```
But, If query starting without whitespace, `select count(*) from user where user.age > :age `

I found that org.springframework.data.jpa.repository.query.QueryUtils.COUNT_MATCH regex not matched select query starting whitespace. I think that it should be fix.

gregturn:
While the example code you provided is SQL, not JPQL, I took your suggestion and added support for count queries starting with whitespace.

The query you wrote will only work if you turn on the nativeQuery flag.

gregturn:
And given that handling whitespace before a count prefix isn't really a bug, I've relabeled this issue as an enhancement, and backported it to the previously supported versions.

Thanks for elevating this @jaoromi!

PR #2394, opened 2021-12-31
GH-2393 The count query pattern matches queries that start with a space.

jaoromi:
The COUNT_MATCH did not consider a query string staring with whitespace. This lead to cause illegal query syntax in the construction of the count query. With the fix we now use a native query string starting with whitespaces with paging.
Related tickets #2393

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -140,6 +140,7 @@ public abstract class QueryUtils {
 		ALIAS_MATCH = compile(builder.toString(), CASE_INSENSITIVE);
 
 		builder = new StringBuilder();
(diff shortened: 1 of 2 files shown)


gregturn:
In light of https://github.com/spring-projects/spring-data-jpa/commit/2545e6d74a550d748c4b59a7bf0972ffc7d547bd, this PR appears to need more work.

gregturn:
We also recently moved branches such that `main` is based upon Spring Data JPA 3.x and Java 17, so you may need to adjust your development environment. I recommend you rebase against `main` to make for a cleaner PR.

gregturn:
I took this PR as inspiration to enhance `QueryUtils` to allow whitespace before count queries.

I must point out, though, that `@Query` is meant for JPQL queries. To do SQL queries, you must switch on `nativeQuery=true`.

Resolved via 94e84cc8.

PR #2398, opened 2022-01-04
Introduce getReferenceById.

### LINKED ISSUE #2232: JpaRepository.getOne/getById method name should indicate that it returns a reference
JpaRepository methods getOne/getById methods are often used by accident by inexperienced developers instead of findById.
These methods call EntityManager::getReference and return a reference.
The method name should indicate that.
We should consider adding method JpaRepository::getReferenceById and deprecating getById (getOne is already deprecated).

gregturn:
Closes #2232

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa</artifactId>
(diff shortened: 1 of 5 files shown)


schauder:
Thanks. That's polished and merged.

PR #2399, opened 2022-01-04
Properly handle Sort.Order for a select field alias.

gregturn:


schauder:
Thanks, this is polished, merged and backported.

PR #2401, opened 2022-01-07
Ignore fields containing Optional.empty when performing Query by Example.

gregturn:


schauder (changes requested):
Can we make the test independent of auditable? Because auditables are just a special case where this problem surfaces.

Also can we have some documentation for this?

gregturn:
Will do.

gregturn:
@schauder I've updated this PR with a non-Auditable based test class.

schauder:
Thanks, that's merged.

PR #2433, opened 2022-02-08
Remove of useless `throws ...` in jUnit tests.

DiegoKrupitza:
A lot of jUnit tests had in the method header a `throws` that is useless. For example `throws Exception` on a method where no exception is thrown. Removing this noice will make the tests more readable.


schauder:
Thanks, that's merged.

PR #2434, opened 2022-02-08
Added query fetches search for join existence when applying JPA ordering.

### LINKED ISSUE #2253: Specifications with sort creates additional join even though the entity was already fetched
Hi, specifications with sort creates additional join even though the entity was already fetched. The result is that the table is then left joined twice. Consider setup:
entities 
```
@Entity
public class A {
    @Id
    Long id

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn(name = "b_id")
    B b;
}

@Entity
public class B {
    @Id
    Long id
    String name
}
```
specifications
```
Specification<A> specification =
(root, query, builder) -> {
   query.distinct(true); 
   root.fetch(A_.b, JoinType.LEFT);
   return builder.equal(root.get(A_.id), 1L);
};

Sort sort = Sort.by(Sort.Direction.ASC, "b.name");

aRepository.findAll(specification, sort);
```
then
`List<T> findAll(Specification<T> spec, Sort sort);`

creates SQL query
```
select distinct A1.id, A1.b_id, B2.id, B2.name
from A A1
    left join B B1 on A1.b_id = B1.id
    left join B B2 on A1.b_id = B2.id
where A.id = 1
order by B1.name asc
```
B entity is joined twice. If the query is distinct, then it fails with SQL error
`Expression #1 of ORDER BY clause is not in SELECT list, references column ... B1.name ... which is not in SELECT list; this is incompatible with DISTINCT
`
The SQL error in this case is correct, order by clause is not in select list. The sort should not create additional join if the entity was already fetched or joined. 

The commit that introduced this behaviour is https://github.com/spring-projects/spring-data-jpa/commit/43d1438b38747addc548bd69f82df0f01818c983

Previously, it checked whether there already is a fetched entity (isAlreadyFetched() method) if yes then it didn’t create additional join (getOrCreateJoin() method). The logic was changed to check whether the entity was already inner-joined. But even if I change the join type in specification to JoinType.INNER (root.fetch(A_.b, JoinType.INNER);)  it still creates additional join. 

Possible duplicate of #2206

AntonMolganov:
Closes #2253
See https://github.com/spring-projects/spring-data-jpa/issues/2253

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -76,6 +76,7 @@ import org.springframework.util.StringUtils;
  * @author Greg Turnquist
  * @author Diego Krupitza
  * @author Jędrzej Biedrzycki
+ * @author Anton Molganov
  */
 public abstract class QueryUtils {
 
@@ -806,6 +807,16 @@ public abstract class QueryUtils {
 				return join;
 			}
 		}
+
+		for (Fetch<?, ?> fetch : from.getFetches()) {
+
+			Join<?, ?> join = (Join<?, ?>) fetch;
+			if (joinType == join.getJoinType()
+					&& join.getAttribute().getName().equals(attribute)) {
+				return join;
+			}
+		}
+
 		return from.join(attribute, joinType);
 	}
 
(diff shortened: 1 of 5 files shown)


schauder:
The "See" link 404s, should that go to #2253?

AntonMolganov:
> The "See" link 404s, should that go to #2253?

Fixed pull request message.
Was .../spring-data-build/... insteadof .../spring-data-jpa/...

schauder (changes requested):
We would need an integration test to go with this. `UserRepositoryTests` would probably a good home for that.

schauder:
This is still missing tests, and should have been automatically closed a long time ago.

PR #2459, opened 2022-02-28
Explicit type replaced with diamond operator.

DiegoKrupitza:
In modern java you do not need to add the explicit type if it can be inferred. To make the code more readable, we removed the explicit type and replaced it with the diamond operator (<>).


schauder:
Thanks. Merged and Backported.

It looks like your formatter is still not configured quite right. I had to undo some format changes.

Issue #2511, opened 2022-04-27
count query issue in spring-data-jpa-2.5.11


lavanya2290:
When we check our native query an issue comes 
the count uery changes from select count(*) from tablename to select count(order) from tablename.

https://github.com/spring-projects/spring-data-jpa/compare/2.5.9...2.5.11#diff-7d36425745042f8c27507657de795fd7cf1d546da6a740fd00d86add46f81ed4R530

lavanya2290:
<img width="1536" alt="Screenshot 2022-04-27 at 4 30 48 PM" src="https://user-images.githubusercontent.com/13776413/165504460-35cef066-8d00-4b56-9dd2-30247e14bd4b.png">
<img width="1536" alt="Screenshot 2022-04-27 at 4 30 40 PM" src="https://user-images.githubusercontent.com/13776413/165504484-9af88631-a5f0-4c70-948a-76451bb1b268.png">

lavanya2290:
Matcher matcher = COUNT_MATCH.matcher("SELECT * FROM context ORDER BY time");
		String countQuery;

		if (countProjection == null) {

			String variable = matcher.matches() ? matcher.group(VARIABLE_NAME_GROUP_INDEX) : null;
			boolean useVariable = StringUtils.hasText(variable) //
					&& !variable.startsWith(" new") //
					&& !variable.startsWith("count(") //
					&& !variable.contains(",") //
					&& !variable.contains("*");

			String complexCountValue = matcher.matches() && StringUtils.hasText(matcher.group(COMPLEX_COUNT_FIRST_INDEX))
					? COMPLEX_COUNT_VALUE
					: COMPLEX_COUNT_LAST_VALUE;

			String replacement = useVariable ? SIMPLE_COUNT_VALUE : complexCountValue;
			countQuery = matcher.replaceFirst(String.format(COUNT_REPLACEMENT_TEMPLATE, replacement));

lavanya2290:
in 2.5.13 spring boot which uses 2.5.11 issue is there.

Axinet:
#2495 is still open, so the only option as for now is downgrade to spring-data-jpa 2.5.10

gregturn:
Support for `2.5.x` ended back in April. I'm closing this ticket for now. If you are still running into this issue on `2.6.x`+, then please open a new ticket with details.

Axinet:
[...]

Issue #2260, opened 2021-07-14
NativeQuery with Pagination


Johnczek:
Hello, I am facing issue similiar with [https://github.com/spring-projects/spring-data-jpa/issues/1282](https://github.com/spring-projects/spring-data-jpa/issues/1282).

I am trying to run this native query with spring-data on version `2.5.0`

`@Query(value = "SELECT dense_rank() OVER (ORDER BY rank DESC)," +
        "t.id, " +
        "FROM table t WHERE t.x > 0 ",
        nativeQuery = true)
Page<Object[]> getT(Pageable pageable);`

But I am getting SQLGrammarException because the spring-data engine will translate this query into

`SELECT dense_rank() OVER (ORDER BY rank DESC),t.id FROM table t WHERE t.x > 0 , t.y desc limit 20`

Where the problem is pretty obvious. There is ORDER BY clause missing, probably because of presence of ORDER BY clause in OVER subquery. Is this a known bug? And is it possible to bypass this somehow? I need to use whole pageable object with native query with this dense_rank() calling.

schauder:
I suspect this is a bug in the underlying JPA implementation. Spring Data JPA uses `Query.setFirstResult` and `Query.setMaxResult` to create the queries for pagination.

Could you confirm that you seeing this behaviour as well if you use an `EntityManager` directly, as described in https://www.baeldung.com/jpa-pagination

Johnczek:
> I suspect this is a bug in the underlying JPA implementation. Spring Data JPA uses `Query.setFirstResult` and `Query.setMaxResult` to create the queries for pagination.
> 
> Could you confirm that you seeing this behaviour as well if you use an `EntityManager` directly, as described in https://www.baeldung.com/jpa-pagination

Hello, I´ve tried many ways but entity manager did not worked for me cause native query has no option to pass sort into query.

But after some investigation I´ve found a workaround with adding ORDER 1 at the end of query and now It is working fine.
I assume that spring will look for ORDER BY substring and if its found, there is (from his point of view) no need to add another ORDER BY and therefore spring appends just , ordering limiting offsetting. But in this case the ORDER BY clause should be added because the found ORDER BY property is not the one that framework is looking for.

TAndronicus:
I'll try to tackle this issue

shu1997:
Hello I am using postgreSQL also facing the same issue-> in my service layer i have created the pageable object using 
inside @Service:-

[...]

PR #2514, opened 2022-04-28
Fixes `*` bug in `createCountQueryFor`.

DiegoKrupitza:
In commit 3e64d9ad9b7d45fcf1231dbaf207be49cc481e7a a bug got introduced that uses the next symbol after the table name for the count function. With this commit this should be now resolved. The count query will use `*` when there is no alias present nor a variable.

Related tickets #2177, #2260, #2511


gregturn:
Backported to 2.7.x.

gregturn:
Backported to 2.6.x.

gregturn:
Backported to 2.5.x

lavanya2290:
which version of 2.5.x is it backported

PR #2521, opened 2022-05-05
Fix case where the from clause is misidentified.

### LINKED ISSUE #2260: NativeQuery with Pagination

darinmanica:
Fix case where the from clause is misidentified.
fixes #2260

Added a zero-width word boundary to the regex that identifies the from clause.  This is used in alias detection for order by clauses.

- [✓] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [✓] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [✓] You submit test cases (unit or integration tests) that back your changes.
- [✓] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -138,7 +138,7 @@ public abstract class QueryUtils {
 	static {
 
 		StringBuilder builder = new StringBuilder();
-		builder.append("(?<=from)"); // from as starting delimiter
+		builder.append("(?<=\\bfrom)"); // from as starting delimiter
 		builder.append("(?:\\s)+"); // at least one space separating
 		builder.append(IDENTIFIER_GROUP); // Entity name, can be qualified (any
 		builder.append("(?:\\sas)*"); // exclude possible "as" keyword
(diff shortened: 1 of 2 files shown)


gregturn:
Thanks @darinmanica! 

Your solution was much simpler than mine. 

Future tip: If you could submit PRs against `main` instead of `2.6.x`, it makes the process to review and then backport to supported branches much easier.

darinmanica:
Thank you @gregturn!  I'll be sure to submit PRs against `main` going forward.

PR #2529, opened 2022-05-11
Make QueryParameterSetter.QueryMetadataCache implementation thread safe.

missingdays:
Previously its `cache` value was set inside the `synchronized` block, but was read without any synchronization. That meant that other threads could read old values of `cache` and compute `QueryMetadata` again for the same key.


gregturn:
The risk, if this isn't accepted, is that we may have to repeat some computations.

The risk, if it IS accepted, is that ALL computations will be slowed down with synchronization efforts, regardless of whether or not the synchronization is needed.

Hence, I don't think this is the solution we need. The current approach is a best-effort approach.

Issue #2540, opened 2022-05-23
HibernateJpaParametersParameterAccessor constructor may throw IllegalStateException


cigaly:
When EntityManagerFactory is created with org.springframework.orm.hibernate5.LocalSessionFactoryBean,
and EntityManager is created with org.springframework.orm.jpa.SharedEntityManagerCreator.createSharedEntityManager,
and HibernateJpaParametersParameterAccessor constructor is invoked out of transaction
then java.lang.IllegalStateException is thrown with message "No transactional EntityManager available"

However, when constructor is invoked inside transaction, no exception is thrown.

Suggested patch that fixes this issue is attached.
[patch.diff.txt](https://github.com/spring-projects/spring-data-jpa/files/8753474/patch.diff.txt)

PR #2542, opened 2022-05-23
Suggested patch to  #2540

cigaly:
Main branch change to issue described in [HibernateJpaParametersParameterAccessor constructor may throw IllegalStateException #2540 ](https://github.com/spring-projects/spring-data-jpa/issues/2540)

Similar change (only difference is jakarta vs javax)should be applied to branch 2.7.x


schauder:
Thanks @cigaly for the PR. Could you share your full name, so we can reference you with an `@author` tag in the source?

cigaly:
> Thanks @cigaly for the PR. Could you share your full name, so we can reference you with an `@author` tag in the source?

Sorry, I've considered that as minor change so did not added author tag. Full name is Cedomir Igaly

schauder:
Thanks, that's merged and back ported.

Issue #1854, opened 2019-06-26
SimpleJpaRepository.delete does not respect JPA `@Access` rules [DATAJPA-1565]


spring-projects-issues:
**[Leonard Brünings](https://jira.spring.io/secure/ViewProfile.jspa?name=leonard84)** opened **[DATAJPA-1565](https://jira.spring.io/browse/DATAJPA-1565?redirect=false)** and commented

With the changes of [DATAJPA-1535](https://jira.spring.io/browse/DATAJPA-1535) (https://github.com/spring-projects/spring-data-jpa/commit/c1d99ae77ed5b69b6955f075b777538ae1e3812d) a bug was introduced.

According to JPA rules the position of the annotation defines whether field or property access is used. It can be overridden with the `@Access` annotation. It seems that org.springframework.data.jpa.repository.support.JpaEntityInformation does not respect that since `entityInformation.getId(entity)` uses `DirectFieldAccessFallbackBeanWrapper` which tries the property access first. This can lead to problems if the getter returns a different type than the type of the actual field.

```java
@Entity
public class ExampleEntity {

    @Id
    @Column(columnDefinition = "uuid")
    @GeneratedValue(generator = "uuid2")
    @GenericGenerator(name = "uuid2", strategy = "uuid2")
    private UUID clientId;

    public UUID getId() {
        return clientId;
    }

    public void setClientId(UUID clientId) {
        this.clientId = clientId;
    }

    public String getClientId() {
        return clientId == null ? null : clientId.toString();
    }
}

```

Fails with (see attached sample)

```
org.springframework.dao.InvalidDataAccessApiUsageException: Provided id of the wrong type for class com.example.datajpaidaccessbug.db.ExampleEntity. Expected: class java.util.UUID, got class java.lang.String; nested exception is java.lang.IllegalArgumentException: Provided id of the wrong type for class com.example.datajpaidaccessbug.db.ExampleEntity. Expected: class java.util.UUID, got class java.lang.String
 at com.example.datajpaidaccessbug.DatajpaIdAccessBugTests.deletionFails(DatajpaIdAccessBugTests.java:22)
```

---

**Affects:** 2.1.9 (Lovelace SR9)

**Attachments:**
- [datajpa-id-access-bug.zip](https://jira.spring.io/secure/attachment/27358/datajpa-id-access-bug.zip) (_54.76 kB_)

**Issue Links:**
[...]

PR #2567, opened 2022-06-13
Refine to use JPA's built in getID feature.

gregturn:
@odrotbohm I'd appreciate your feedback on whether or not my proposal makes sense.

This change has ACTUALLY allowed me to enable some test cases that were disabled before. And it also addresses issues like https://github.com/spring-projects/spring-data-jpa/issues/1854.

The one part I'm really uncertain about, is whether to use `DirectFieldAccessFallbackBeanWrapper` the way I have it coded in this commit, or Thomas Darimont's `IdentifierDerivingDirectFieldAccessFallbackBeanWrapper` extension that seemingly incorporates more "JPA knowledge" or not.

Thoughts?


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JpaMetamodelEntityInformation.java
```diff
@@ -51,26 +50,28 @@
...
 	/**
 	 * Creates a new {@link JpaMetamodelEntityInformation} for the given domain class and {@link Metamodel}.
-	 *
+	 * 
 	 * @param domainClass must not be {@literal null}.
-	 * @param metamodel must not be {@literal null}.
+	 * @param entityManager#metamodel must not be {@literal null}.
 	 */
-	public JpaMetamodelEntityInformation(Class<T> domainClass, Metamodel metamodel) {
+	public JpaMetamodelEntityInformation(Class<T> domainClass, EntityManager entityManager) {
```
odrotbohm, line 69: Does it make sense to rather change the constructor to just add `PersistenceUnitUtil` as an additional parameter to avoid the need to depend on an `EntityManager` and move the preparation of the parameters to `JpaEntityInformationSupport `? This should also ease testing as we don't have to mock the intermediate mocking steps.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
```diff
@@ -184,7 +184,8 @@ public void delete(T entity) {
 
 		Class<?> type = ProxyUtils.getUserClass(entity);
 
-		T existing = (T) em.find(type, entityInformation.getId(entity));
+		Object id = em.getEntityManagerFactory().getPersistenceUnitUtil().getIdentifier(entity);
```
odrotbohm, line 187: Is that a left-over from initial experiments? We should be able to still call `entityInformation.getId(…)` right?

odrotbohm (changes requested):
LGTM, except the comments on the constructor of `Jpa…EntityInformation`. If you agree with them and make the changes accordingly, feel free to merge with those changes included.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
gregturn: Yes it is. Thanks!

PR #2557, opened 2022-06-01
Handle newlines in subquery detection

freafrea:
- [V] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [V] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [V] You submit test cases (unit or integration tests) that back your changes.
- [V] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).
- 

Added support for handling newlines in finding of subqueries. Changed . to match newlines in regex.

PR #2582, opened 2022-06-28
Handle multiline subquery removal

c-fraser:


freafrea:
Hi, this looks like almost the same change as the one I've added previously :)
https://github.com/spring-projects/spring-data-jpa/pull/2557
Do you know of any way of speeding up the acceptance? Already we have both reimplemented the same behaviour.

gregturn:
Backported to `2.7.x.` and `2.6.x`.

Issue #2594, opened 2022-07-14
limit queries of single finders to 2


tschniewind:
**Enhancement:**
Given a repositorys find-method returning a single result or an java.util.Optional should **limit** the underlying query to 2.

This would help to get the expected Result in case of multiple rows (IncorrectResultSizeDataAccessException) even if the query leads to huge result set.

(We had a case leading to an out of memory exception cause the query found about 50000 rows. Unfortunately we could not use a UniqueConstraint.)

tschniewind:
In which version will this enhancement be integrated, any idea?

schauder:
This will be part of the next 3.0 milestone and therefore eventually in 3.0 GA.

gregturn:
Backported to `2.7.x`

gregturn:
Backported to `2.6.x`.

PR #2604, opened 2022-07-26
Limit single finders max results to 2 for performance

quaff:
Close GH-2594


schauder:
Thanks, that is merged.

PR #2609, opened 2022-08-01
Support Contains with ElementCollection of type String.

### LINKED ISSUE #2607: contains in query derivation of an ElementCollection creates a proper isMember clause but encloses the argument in % leading to the wrong result.
When looking at the code `CONTAINING` is obviously intended to work with Strings (and be translated to  `x like %?%`) and for collections where it gets translated to an `isMemberOf`. But in the later case the argument is still wrapped with `%` resulting in the condition yielding the wrong results. In most cases non at all. 

Interesting using a simple property does work.

schauder:
Part of the logic considered `contains` on an ElementCollection of type String a like query, wrapping the parameter in wildcards.
This is now fixed.

Closes #2607

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
@@ -196,6 +196,7 @@ class ParameterMetadataProvider {
 		private final ParameterExpression<T> expression;
 		private final EscapeCharacter escape;
 		private final boolean ignoreCase;
+		private final boolean noWildcards;
 
 		/**
 		 * Creates a new {@link ParameterMetadata}.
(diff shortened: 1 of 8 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java
```diff
@@ -241,7 +243,7 @@ public Object prepare(Object value) {
 				return value;
 			}
 
-			if (String.class.equals(expressionType)) {
+			if (String.class.equals(expressionType) && !noWildcards) {
```
v-ladynev, line 246: Better to create a separate method for this
```java
private boolean aMethod() {
    return String.class.equals(expressionType) && !noWildcards;
}
```
 and say in the method name what is actually checked here.

PR #2624, opened 2022-08-26
Fix typo on AuditingEntityListener ('persiting' on line 29)

furnivall:
Just a small change to fix a typo within the docs.


KianKwa (approved):

spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/support/AuditingEntityListener.java
```diff
@@ -26,7 +26,7 @@
 import org.springframework.util.Assert;
 
 /**
- * JPA entity listener to capture auditing information on persiting and updating entities. To get this one flying be
+ * JPA entity listener to capture auditing information on persisting and updating entities. To get this one flying be
```
KianKwa, line 29: Good spot Danny!

Issue #2641, opened 2022-09-23
SqlParserQueryEnhancer errors with Merge statements


xercasy:
Hello

similar to #2593

With spring boot >= 2.7, the application failed to start because of merge statement (oracle syntax)

> class net.sf.jsqlparser.statement.merge.Merge cannot be cast to class net.sf.jsqlparser.statement.select.Select (net.sf.jsqlparser.statement.merge.Merge and net.sf.jsqlparser.statement.select.Select are in unnamed module of loader 'app')

for exemple in a JPA repository

```
@Modifying
@Query(value = "merge into ...", nativeQuery = true)
void mergeStatement();
```

`detectParsedType()` in _JSqlParserQueryEnhancer_ appears to neglect the Merge statement which is supported by Jsqlparser

gregturn:
@DiegoKrupitza Is this a scenario you're familiar with?

DiegoKrupitza:
> @DiegoKrupitza Is this a scenario you're familiar with?

I will take a look at this! I have seen that @xercasy has already proposed a PR for this (#2642).

DiegoKrupitza:
@gregturn As realized thanks to the issue #2593 I did not implement all the possible Statements JSqlParser can handle. In #2595 I added support for `INSERT`. Thanks to @xercasy the PR #2642 will fix it with `MERGE`. 

I already reviewed the PR #2642 and as far as I can see it should fix the problem.

gregturn:
Thanks to @xercasy and @DiegoKrupitza, this additional support for native `MERGE` operations is now merged to `main`.

xercasy:
@gregturn Can you backport this issue and https://github.com/spring-projects/spring-data-jpa/issues/2593 to 2.7.x ? Thanks

gregturn:
Backported to `2.7.x`.

PR #2642, opened 2022-09-23
Made JSqlParserQueryEnhancer aware of MERGE statements.

xercasy:
Related tickets https://github.com/spring-projects/spring-data-jpa/issues/2641

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancer.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancer.java
@@ -18,27 +18,6 @@ package org.springframework.data.jpa.repository.query;
 import static org.springframework.data.jpa.repository.query.JSqlParserUtils.*;
 import static org.springframework.data.jpa.repository.query.QueryUtils.*;
 
-import net.sf.jsqlparser.JSQLParserException;
-import net.sf.jsqlparser.expression.Alias;
-import net.sf.jsqlparser.expression.Expression;
-import net.sf.jsqlparser.expression.Function;
-import net.sf.jsqlparser.parser.CCJSqlParserUtil;
-import net.sf.jsqlparser.schema.Column;
-import net.sf.jsqlparser.statement.Statement;
-import net.sf.jsqlparser.statement.delete.Delete;
-import net.sf.jsqlparser.statement.merge.Merge;
-import net.sf.jsqlparser.statement.insert.Insert;
-import net.sf.jsqlparser.statement.select.OrderByElement;
-import net.sf.jsqlparser.statement.select.PlainSelect;
-import net.sf.jsqlparser.statement.select.Select;
-import net.sf.jsqlparser.statement.select.SelectBody;
-import net.sf.jsqlparser.statement.select.SelectExpressionItem;
-import net.sf.jsqlparser.statement.select.SelectItem;
-import net.sf.jsqlparser.statement.select.SetOperationList;
-import net.sf.jsqlparser.statement.select.WithItem;
-import net.sf.jsqlparser.statement.update.Update;
-import net.sf.jsqlparser.statement.values.ValuesStatement;
-
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancer.java
```diff
@@ -478,7 +482,7 @@ public DeclaredQuery getQuery() {
 	 * </ul>
 	 */
```
DiegoKrupitza, line 483: Please add the new `MERGE` to the JavaDoc of the enum.

`<li>{@code ParsedType.MERGE}: means the top level statement is {@link Merge}</li>`

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryEnhancerUnitTests.java
```diff
@@ -889,6 +890,18 @@ void multipleWithStatementsWorksWithJSQLParser() {
...
 	}
 
+	@Test // GH-2641
+	void mergeStatementWorksWithJSqlParser() {
+		String query = "merge into a using (select id, value from b) query on (a.id = query.id) when matched then update set a.value = value";
+		StringQuery stringQuery = new StringQuery(query, true);
+		QueryEnhancer queryEnhancer = QueryEnhancerFactory.forQuery(stringQuery);
+
+		assertThat(queryEnhancer.getJoinAliases()).isEmpty();
+		assertThat(queryEnhancer.detectAlias()).isNull();
```
DiegoKrupitza, line 900: I don't know if this is ever needed but as far as I can see a merge has an "alias" (see screenshot). 
Maybe it should not return `NULL` here. Is it possible to adapt the `detectAlias()` to consider this edge case without making the method overly complicated? 

image

DiegoKrupitza (changes requested):
I like your extension :) It should resolve the problem. Please review my comments. The JavaDoc is important, so the documentation stays up-to-date. Regarding the alias I would love to hear your take on this.

DiegoKrupitza:
I have seen this is your first PR here, welcome to Open Source 🥳👍  While reviewing your PR I have seen your forgot to add the `Closes #2641` to your commit. Adding this is quite helpful because then GitHub then create automatic links between commits and issues. 

https://github.com/spring-projects/spring-data-build/blob/main/CONTRIBUTING.adoc#commit-messages

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryEnhancerUnitTests.java
xercasy: I tried to adapt the `detectAlias()` method for JSqlParserQueryEnhancer class, however it is more difficult for me to modify the method for the `DefaultQueryEnhancer` ([82590bf](https://github.com/spring-projects/spring-data-jpa/pull/2642/commits/82590bf25b5b6281da6e0844f3957f1ef0c507d5))
I am waiting your feedback

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancer.java
```diff
@@ -342,6 +346,18 @@ private static String detectAlias(PlainSelect selectBody) {
...
 	}
 
+	/**
+	 * Resolves the alias for the given {@link Merge} statement.
+	 *
+	 * @param mergeStatement must not be {@literal null}.
+	 * @return Might return {@literal null}.
+	 */
+	@Nullable
+	private static String detectAlias(Merge mergeStatement) {
```
DiegoKrupitza, line 356: Is there a reason why you decided to make the method static?

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancer.java
```diff
@@ -449,10 +465,10 @@ public Set<String> getJoinAliases() {
 	 * @param query the query to parse
 	 * @return the parsed query
 	 */
-	private static Select parseSelectStatement(String query) {
+	private static <T extends Statement> T parseSelectStatement(String query, Class<T> classOfT) {
```
DiegoKrupitza, line 468: Thats a clever solution! Make it more flexible in the future.
Personally I would also add a "default" version where we know it will be a `Select`. Because as you can see above we often call `parseSelectStatement(query.getQueryString(), Select.class);`.

```java
       /**
	 * Parses a query string with JSqlParser.
	 *
	 * @param query the query to parse
	 * @return the parsed query
	 */
	private Select parseSelectStatement(String query) {
		return parseSelectStatement(query,Select.class);
	}
```

DiegoKrupitza (commented):
All in all it looks good. I just left two comments just for personal taste.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancer.java
xercasy: I just kept the same behavior for the function just above (static method) to have a homogeneous code :
private static String detectAlias(PlainSelect selectBody)

xercasy: I made a force push ... I hope this is not a mistake

gregturn:
Thanks to @xercasy and @DiegoKrupitza! I have polished and merged this PR to `main`.

Issue #2348, opened 2021-11-04
count query fails to be auto generated when using @Query annotation with fetch join


sinbom:
findAllFetch method causes the following error. however, findAllNotFetch method does not cause errors. 
because the automatically generated countQuery does not include the fetch keyword. likewise, fetch join paging using @EntityGraph does not cause an error.

```java
public interface BoardRepository extends JpaRepository<Board, Long> {

    @Query(value = "SELECT DISTINCT b FROM Board b LEFT JOIN FETCH b.comments ORDER BY b.id")
    Page<Board> findAllFetch(Pageable pageable);
    
    @Query(value = "SELECT DISTINCT b FROM Board b LEFT JOIN b.comments ORDER BY b.id")
    Page<Board> findAllNotFetch(Pageable pageable);

    @EntityGraph(attributePaths = "comments")
    Page<Board> findAll(Pageable pageable);

}
```

```
Caused by: org.hibernate.QueryException: query specified join fetching, but the owner of the fetched association was not present in the select list [FromElement{explicit,not a collection join,fetch join,fetch non-lazy properties,classAlias=null,role=com.example.demo.entity.Board.comments,tableName=comment,tableAlias=comments1_,origin=board board0_,columns={board0_.id,className=com.example.demo.entity.Comment}}]
	at org.hibernate.hql.internal.ast.tree.SelectClause.initializeExplicitSelectClause(SelectClause.java:215) ~[hibernate-core-5.4.32.Final.jar:5.4.32.Final]
	at org.hibernate.hql.internal.ast.HqlSqlWalker.useSelectClause(HqlSqlWalker.java:1028) ~[hibernate-core-5.4.32.Final.jar:5.4.32.Final]
	at org.hibernate.hql.internal.ast.HqlSqlWalker.processQuery(HqlSqlWalker.java:796) ~[hibernate-core-5.4.32.Final.jar:5.4.32.Final]
	at org.hibernate.hql.internal.antlr.HqlSqlBaseWalker.query(HqlSqlBaseWalker.java:694) ~[hibernate-core-5.4.32.Final.jar:5.4.32.Final]
	at org.hibernate.hql.internal.antlr.HqlSqlBaseWalker.selectStatement(HqlSqlBaseWalker.java:330) ~[hibernate-core-5.4.32.Final.jar:5.4.32.Final]
[...]

PR #2651, opened 2022-10-02
Remove 'fetch' part in 'join fetch' clause during generation of count query

yukharin:
Closes https://github.com/spring-projects/spring-data-jpa/issues/2348

- [+] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [+] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [+] You submit test cases (unit or integration tests) that back your changes.
- [+] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).


yukharin:
Previously, I was told not to include merge commits in PR.
In this PR, it's done, but right now, I am not sure about rebasing my PR on main and adding main's commits to my current PR. 
If it's not ok, please, let me know) 

One more thing to add here is that I've tested my fix locally in some real project and it worked fine. I've not added any integration tests to cover my fix, but once again, if you think that it's needed, let me know)

PR #2655, opened 2022-10-03
Unwrapping of TypedParameterValue, return null instead of empty strin…

### LINKED ISSUE #2653: Ambiguous query generation with null-valued parameters in like-queries
### **Problem**
-------------------------
Actually the issue [2548](https://github.com/spring-projects/spring-data-jpa/issues/2548) has not been completely resolved! 
`PersistenceProvider#condense(Object)` returns an empty string, so then the query
```@Query(SELECT t FROM My_Table t WHERE (:param is NULL or t.attr LIKE %:param%))```
will transform into ```select my_table.id, ..., my_table.attr as attrN from my_table where '%%' is null or my_table.attr like '%%'``` whereas in spring-data-jpa 2.6.4 or earlier it will transform to ```select my_table.id, ..., my_table.attr as attrN from my_table where NULL is null or my_table.attr like NULL```

yura2201:
…g for string params.

Fixes #2653
Related tickets #2548

### DIFF
--- a/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
+++ b/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
@@ -50,6 +50,7 @@ import org.springframework.util.ConcurrentReferenceHashMap;
  * @author Thomas Darimont
  * @author Mark Paluch
  * @author Jens Schauder
+ * @author Yuriy Tsarkov
  */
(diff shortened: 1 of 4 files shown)


gregturn:
@yura2201 See if the snapshots now solve this problem.

gregturn:
With #2653 resolved, I'm closing this PR.

PR #2679, opened 2022-10-23
Make @EnableJpaRepositories repeatable

### LINKED ISSUE #2678: Make @EnableJpaRepositories repeatable
In a context where there are multiple EntityManagerFactories / TransactionManagers, being able to declare multiple `@EnableJpaRepositories` on the same class would be helpful.

reda-alaoui:
Fix #2678


gregturn:
Simply asking for an annotation to be repeatedly doesn't suffice. We need an example confirming it acts as expected and also some test cases illustrating the benefit.

reda-alaoui:
This PR was indirectly related to https://github.com/spring-projects/spring-data-jpa/issues/2685 .
Although I don't agree with the way https://github.com/spring-projects/spring-data-jpa/issues/2685 was rejected, the current PR makes no more sense anymore.

Closing this PR because https://github.com/spring-projects/spring-data-jpa/issues/2685 was rejected.

PR #2686, opened 2022-11-01
Remove 'fetch' part in 'join fetch' clause during generation of count query

yukharin:
Closes https://github.com/spring-projects/spring-data-jpa/issues/2348
- [+] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [+] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [+] You submit test cases (unit or integration tests) that back your changes.
- [+] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).


yukharin:
I've tested this fix on some real project and it works fine. 
But if there is a need to add integration tests, please, let me know)

gregturn:
We are working on a parsing solution to better handle such situations than `QueryUtils` accommodates. For the meantime, we are trying to avoid any more changes to `QueryParser`.

gregturn:
Superseded by https://github.com/spring-projects/spring-data-jpa/pull/2883.

PR #2722, opened 2022-12-07
Fix generics for JpaSpecificationExecutor::findBy

quaff:
`Specification<T>` does not hold type `S` like `Example<S>`

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/JpaSpecificationExecutor.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/JpaSpecificationExecutor.java
@@ -32,6 +32,7 @@ import org.springframework.lang.Nullable;
  * @author Oliver Gierke
  * @author Christoph Strobl
  * @author Diego Krupitza
+ * @author Yanming Zhou
  */
 public interface JpaSpecificationExecutor<T> {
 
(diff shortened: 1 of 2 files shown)


gregturn:
Thanks @quaff. I check out your PR and indeed it works. But I think I'll pass on it.

Whereas right now, the `Specification` aspects of `SimpleJpaRepository` (`<S extends T, R> R findBy(Specification<T> spec, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction)`) don't leverage the generic sub-type of `S`, I'd rather maintain the symmetric nature of this API with our Query By Example variant (`<S extends T, R> R findBy(Example<S> example, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction)`), in the event that we do in the future.

I realize that the culprit is `sort -> getQuery(spec, getDomainClass(), sort)` and that is hardwires the current implementation into a `T`, but if we are able to loosen that up in the future, the code will already be in place to handle this and people won't have migrated into a stricter situation that we can't get out of without a major release.

quaff:
> Thanks @quaff. I check out your PR and indeed it works. But I think I'll pass on it.
> 
> Whereas right now, the `Specification` aspects of `SimpleJpaRepository` (`<S extends T, R> R findBy(Specification<T> spec, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction)`) don't leverage the generic sub-type of `S`, I'd rather maintain the symmetric nature of this API with our Query By Example variant (`<S extends T, R> R findBy(Example<S> example, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction)`), in the event that we do in the future.
> 
> I realize that the culprit is `sort -> getQuery(spec, getDomainClass(), sort)` and that is hardwires the current implementation into a `T`, but if we are able to loosen that up in the future, the code will already be in place to handle this and people won't have migrated into a stricter situation that we can't get out of without a major release.

Then you should use diamond operator and return `FetchableFluentQuery<S>` to avoid casting.
```java
		FetchableFluentQuery<S> fluentQuery = new FetchableFluentQueryBySpecification<>(spec, getDomainClass(),
				Sort.unsorted(), null, finder, this::count, this::exists, this.em);

		return queryFunction.apply(fluentQuery);
```

PR #2739, opened 2022-12-22
Change calls to deprecated methods in tests to their replacements.

ErikPelli:
ClassTypeInformation is deprecated and will be removed, use TypeInformation instead. 

StringUtils.trimWhitespace can be replaced with the strip method called on the String object. StandardAnnotationMetadata is deprecated in favor of the factory method AnnotationMetadata.introspect(Class).


gregturn:
Thanks @ErikPelli! I took your patch and did a little polished, then merged to main.

PR #2766, opened 2023-01-17
Fixed order by subwindow regex by making it lazy instead of greedy

### LINKED ISSUE #2764: Spring data jpa QueryUtils applysorting not working for multiple order by statements in seperate window functions
Hello,

I ran into a bug with the Pageable object, and with some deep digging found out a bug in the QueryUtils, when applying the regex of `ORDER_BY_IN_WINDOW_OR_SUBSELECT` on my query. \
I looked at your test suite in the class named `QueryUtilsUnitTests.java` and tried adding a test myself locally with my current query e.g. \
```
	@Test
	void orderByShouldWorkWithMultipleSubSelectStatements() {
		Sort sort = Sort.by(Order.desc("weight"));

		assertThat(QueryUtils.applySorting("""
				WITH topics AS (
				    SELECT      topic.*
				    FROM        objective topic
				    INNER JOIN  user
				    ON          user.id = topic.user_id
				    WHERE       user.aggregate_id = 'someId'
				    AND          CASE
				                    WHEN :__$synthetic$__4 IS NOT NULL THEN topic.last_update_date >= :__$synthetic$__5
				                    ELSE FALSE
				                END)
				),
				 WITH open_topics AS (
				    SELECT  topic.*,
				            row_number() OVER (PARTITION BY topic.user_id ORDER BY topic.target_date DESC, topic.name DESC) AS weight
				    FROM    topics topic
				    WHERE   topic.assessment = 'OPEN'
				),
				 other_topics AS (
				    SELECT  topic.*,
				            row_number() OVER (PARTITION BY topic.user_id ORDER BY topic.target_date DESC, topic.name) AS weight
				    FROM    topics topic
				    WHERE   topic.assessment <> 'OPEN'
				),
				SELECT  topic.*
				FROM    weighed_topics topic
				""", sort)).endsWith("order by weight desc");
	}
```
As I expected, it fails.
```
java.lang.AssertionError: 
Expecting actual:
  "WITH topics AS (
    SELECT      topic.*
    FROM        objective topic
    INNER JOIN  user
    ON          user.id = topic.user_id
    WHERE       user.aggregate_id = 'someId'
    OR          CASE
                    WHEN :__$synthetic$__4 IS NOT NULL THEN topic.last_update_date >= :__$synthetic$__5
                    ELSE FALSE
                END)
),
 WITH open_topics AS (
    SELECT  topic.*,
            row_number() OVER (PARTITION BY topic.user_id ORDER BY topic.target_date DESC, topic.name DESC) AS weight
    FROM    topics topic
    WHERE   topic.assessment = 'OPEN'
),
 other_topics AS (
    SELECT  topic.*,
            row_number() OVER (PARTITION BY topic.user_id ORDER BY topic.target_date DESC, topic.name) AS weight
    FROM    topics topic
    WHERE   topic.assessment <> 'OPEN'
),
SELECT  topic.*
FROM    weighed_topics topic
, weight desc"
to end with:
  "order by weight desc"
```
It is giving me a comma while it actually should be an `order by` statement. This is because the function `hasOrderByClause` in `QueryUtils` eventually expresses itself as `2 > 1`, causing it to use a comma instead of the statement.

```
	private static boolean hasOrderByClause(String query) {
		return countOccurrences(ORDER_BY, query) > countOccurrences(ORDER_BY_IN_WINDOW_OR_SUBSELECT, query);
	}

        ....
	
	if (hasOrderByClause(query)) {
		builder.append(", ");
	} else {
		builder.append(" order by ");
	}
```

As you can see, there actually are 2 `order by` statements in the query, in the PARTITION part. \ 
It seems like the regex expression `ORDER_BY_IN_WINDOW_OR_SUBSELECT` has some issues with multiple `order by` statements in these `WITH` queries. So my guess is that the regex is incorrect and could be adjusted.

brianmarting:
Changed the ORDER_BY_IN_WINDOW_OR_SUBSELECT pattern in the QueryUtils class so that it can handle multiple ORDER BY statements in a window or subselect. Added a test to verify that it works.

Closes #2764

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -105,7 +105,7 @@ public abstract class QueryUtils {
 	private static final String EQUALS_CONDITION_STRING = "%s.%s = :%s";
 	private static final Pattern ORDER_BY = Pattern.compile("(order\\s+by\\s+)", CASE_INSENSITIVE);
 	private static final Pattern ORDER_BY_IN_WINDOW_OR_SUBSELECT = Pattern
-			.compile("\\([\\s\\S]*order\\s+by\\s[\\s\\S]*\\)", CASE_INSENSITIVE);
+			.compile("\\([\\s\\S]*?order\\s+by\\s[\\s\\S]*?\\)", CASE_INSENSITIVE);
 
 	private static final Pattern NAMED_PARAMETER = Pattern.compile(COLON_NO_DOUBLE_COLON + IDENTIFIER + "|#" + IDENTIFIER,
 			CASE_INSENSITIVE);
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryUtilsUnitTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryUtilsUnitTests.java
@@ -48,6 +48,7 @@ import org.springframework.util.StringUtils;
  * @author Darin Manica
  * @author Chris Fraser
  * @author Michał Pachucki
+ * @author Brian Marting
  */
 class QueryUtilsUnitTests {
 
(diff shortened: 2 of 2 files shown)


gregturn:
@brianmarting I'm afraid we can't accept such a widely scoped change to `QueryUtils`. Changing from greedy to lazy is a wide-swinging change that may not impact the existing test cases already here and work for the additional test case submitted, but such a bug change has the potential to break other queries we're not fully aware of.

Instead, the Spring Data team is looking for a deeper, more practical solution rather than making tweaks here both regarding SQL/JPQL parsing as well as ability to tune generated queries, instead of fiddling with regular expressions.

Jayanthreddy:
Since this PR declined, Is there any workaround for this issue. we are facing this issue with our upgrade of all the tech stack which are supportable respective versions of java17.

Thanks

PR #2777, opened 2023-01-25
Differentiate between JPQL and native queries in count query derivation

### LINKED ISSUE #2773: Broken count query for native MySQL queries with table alias
The following repository currently does not work with MySQL:

```java
public interface SomeRepository extends CrudRepository<SomeEntity, Long> {

  @Query(
      value = "SELECT * FROM table_name some_alias",
      nativeQuery = true
  )
  Page<SomeEntity> search(Pageable pageable);

}
```

The repository worked fine with Spring Data JPA 2.6.3, has been broken since 2.6.4, and is still broken in 2.7.7.

The log on failure is

> Hibernate: SELECT * FROM table_name some_alias ?
> Hibernate: select count(some_alias) FROM table_name some_alias
>  WARN --- o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 1054, SQLState: 42S22
> ERROR --- o.h.engine.jdbc.spi.SqlExceptionHelper   : Unknown column 'acc' in 'field list'

The problem is the count query that should be `select count(*) FROM table_name some_alias` with `*` instead of the alias as parameter to `COUNT`.

The regression seems to have been introduced with #2389 as the PR contains a test case for the erroneous behavior.

mp911de:
We now consider whether a query is a native one when deriving a count query for pagination. Previously, the generated queries used JPQL syntax that doesn't comply with native SQL syntax rules.

Also, cleanup tests, deduplicate test code, and rearrange test cases into a maintainable form. 

Closes #2773

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -18,10 +18,23 @@ package org.springframework.data.jpa.repository.query;
 import static jakarta.persistence.metamodel.Attribute.PersistentAttributeType.*;
 import static java.util.regex.Pattern.*;
 
-import jakarta.persistence.*;
-import jakarta.persistence.criteria.*;
-import jakarta.persistence.metamodel.*;
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToOne;
+import jakarta.persistence.Parameter;
+import jakarta.persistence.Query;
+import jakarta.persistence.criteria.CriteriaBuilder;
+import jakarta.persistence.criteria.Expression;
+import jakarta.persistence.criteria.Fetch;
+import jakarta.persistence.criteria.From;
+import jakarta.persistence.criteria.Join;
+import jakarta.persistence.criteria.JoinType;
+import jakarta.persistence.metamodel.Attribute;
 import jakarta.persistence.metamodel.Attribute.PersistentAttributeType;
+import jakarta.persistence.metamodel.Bindable;
+import jakarta.persistence.metamodel.ManagedType;
+import jakarta.persistence.metamodel.PluralAttribute;
+import jakarta.persistence.metamodel.SingularAttribute;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
@@ -570,6 +583,19 @@ public abstract class QueryUtils {
 	 */
 	@Deprecated
 	public static String createCountQueryFor(String originalQuery, @Nullable String countProjection) {
+		return createCountQueryFor(originalQuery, countProjection, false);
+	}
+
+	/**
+	 * Creates a count projected query from the given original query.
+	 *
+	 * @param originalQuery must not be {@literal null}.
+	 * @param countProjection may be {@literal null}.
+	 * @param nativeQuery whether the underlying query is a native query.
+	 * @return a query String to be used a count query for pagination. Guaranteed to be not {@literal null}.
+	 * @since 2.7.8
+	 */
+	static String createCountQueryFor(String originalQuery, @Nullable String countProjection, boolean nativeQuery) {
 
 		Assert.hasText(originalQuery, "OriginalQuery must not be null or empty");
 
@@ -591,9 +617,14 @@ public abstract class QueryUtils {
 
 			String replacement = useVariable ? SIMPLE_COUNT_VALUE : complexCountValue;
 
-			String alias = QueryUtils.detectAlias(originalQuery);
-			if ("*".equals(variable) && alias != null) {
-				replacement = alias;
+			if (nativeQuery && (variable.contains(",") || "*".equals(variable))) {
+				replacement = "1";
+			} else {
+
+				String alias = QueryUtils.detectAlias(originalQuery);
+				if (("*".equals(variable) && alias != null)) {
+					replacement = alias;
+				}
 			}
 
 			countQuery = matcher.replaceFirst(String.format(COUNT_REPLACEMENT_TEMPLATE, replacement));
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancerUnitTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancerUnitTests.java
@@ -15,10 +15,23 @@
  */
 package org.springframework.data.jpa.repository.query;
 
+import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assumptions.*;
+
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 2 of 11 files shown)


spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancerUnitTests.java
```diff
@@ -27,4 +40,181 @@ QueryEnhancer createQueryEnhancer(DeclaredQuery declaredQuery) {
...
+		assertThat(queryEnhancer.createCountQueryFor()).isEqualToIgnoringCase(setQuery);
+		assertThat(queryEnhancer.applySorting(Sort.by("SOME_COLUMN").ascending())).endsWith("ORDER BY SOME_COLUMN ASC");
+		assertThat(queryEnhancer.getJoinAliases()).isEmpty();
+		assertThat(queryEnhancer.detectAlias()).isNullOrEmpty();
+		assertThat(queryEnhancer.getProjection()).isEqualToIgnoringCase("SOME_COLUMN");
+		assertThat(queryEnhancer.hasConstructorExpression()).isFalse();
+	}
+
+	@Test // GH-2578
+	void deeplyNestedcomplexSetOperationListWorks() {
```
gregturn, line 102: Perhaps make it `deeplyNestedComplexSetOperationListWorks`?

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancerUnitTests.java
```diff
@@ -27,4 +40,181 @@ QueryEnhancer createQueryEnhancer(DeclaredQuery declaredQuery) {
...
+		assertThat(queryEnhancer.createCountQueryFor()).isEqualToIgnoringCase(setQuery);
+		assertThat(queryEnhancer.applySorting(Sort.by("CustomerID").descending())).endsWith("ORDER BY CustomerID DESC");
+		assertThat(queryEnhancer.getJoinAliases()).isEmpty();
+		assertThat(queryEnhancer.detectAlias()).isNullOrEmpty();
+		assertThat(queryEnhancer.getProjection()).isEqualToIgnoringCase("CustomerID");
+		assertThat(queryEnhancer.hasConstructorExpression()).isFalse();
+	}
+
+	@Test // GH-2578
+	void valuesStatementsWorks() {
```
gregturn, line 129: Perhaps `valuesStatementWorks`?

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancerUnitTests.java
```diff
@@ -27,4 +40,181 @@ QueryEnhancer createQueryEnhancer(DeclaredQuery declaredQuery) {
...
+		assertThat(queryEnhancer.createCountQueryFor()).isEqualToIgnoringCase(setQuery);
+		assertThat(queryEnhancer.applySorting(Sort.by("CustomerID").descending())).isEqualTo(setQuery);
+		assertThat(queryEnhancer.getJoinAliases()).isEmpty();
+		assertThat(queryEnhancer.detectAlias()).isNullOrEmpty();
+		assertThat(queryEnhancer.getProjection()).isNullOrEmpty();
+		assertThat(queryEnhancer.hasConstructorExpression()).isFalse();
+	}
+
+	@Test // GH-2578
+	void withStatementsWorks() {
```
gregturn, line 149: Perhaps `withStatementWorks`?

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JSqlParserQueryEnhancerUnitTests.java
```diff
@@ -27,4 +40,181 @@ QueryEnhancer createQueryEnhancer(DeclaredQuery declaredQuery) {
...
+						+ "SELECT count(1) FROM sample_data AS a");
+		assertThat(queryEnhancer.applySorting(Sort.by("day").descending())).endsWith("ORDER BY a.day DESC");
+		assertThat(queryEnhancer.getJoinAliases()).isEmpty();
+		assertThat(queryEnhancer.detectAlias()).isEqualToIgnoringCase("a");
+		assertThat(queryEnhancer.getProjection()).isEqualToIgnoringCase("day, value");
+		assertThat(queryEnhancer.hasConstructorExpression()).isFalse();
+	}
+
+	@Test // GH-2578
+	void multipleWithStatementsWorks() {
```
gregturn, line 172: Perhaps `multipleWithStatementsWork`?

gregturn (approved):
I put in some naming suggestions. But apart from that, it looks great. Like the TCK tactic.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -570,6 +583,19 @@ public static String createCountQueryFor(String originalQuery) {
...
+	/**
+	 * Creates a count projected query from the given original query.
+	 *
+	 * @param originalQuery must not be {@literal null}.
+	 * @param countProjection may be {@literal null}.
+	 * @param nativeQuery whether the underlying query is a native query.
+	 * @return a query String to be used a count query for pagination. Guaranteed to be not {@literal null}.
+	 * @since 2.7.8
+	 */
+	static String createCountQueryFor(String originalQuery, @Nullable String countProjection, boolean nativeQuery) {
```
schauder, line 598: I'd prefer two separate methods `createCountQueryForNative` and `createCountQueryForJpql` plus extracting the common parts into a new method.

schauder (approved):

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
mp911de: My gut feeling is that we should rethink `QueryUtils` entirely and eventually eliminate the most methods (or create delegates). Therefore, I would want to refrain from introducing additional methods for the time being unless new methods heavily improve the code.

schauder:
That's merged and back ported to 3.0.x

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -591,9 +617,14 @@ public static String createCountQueryFor(String originalQuery, @Nullable String
 
 			String replacement = useVariable ? SIMPLE_COUNT_VALUE : complexCountValue;
 
-			String alias = QueryUtils.detectAlias(originalQuery);
-			if ("*".equals(variable) && alias != null) {
-				replacement = alias;
+			if (nativeQuery && (variable.contains(",") || "*".equals(variable))) {
```
CardoEggert, line 620: What if it is a native query and the variable is null 😬

mp911de: If you run into a bug or you can reproduce such a case, please file a bug report.

CardoEggert: already present https://github.com/spring-projects/spring-data-jpa/issues/2812

gmbeneva:
will it be back ported to 2.7.x?

mp911de:
Yes, we will backport the fix to 2.7.x.

PR #2817, opened 2023-02-23
Fix NullPointer when deriving a count query for non-SELECT statements, delay count query derivation

mp911de:
Should be backported to `2.7.x` and `3.0.x` as well.


gregturn (approved):
Looks good to me!

gregturn:
Fixed in `main` and backported to `3.0.x` and `2.7.x`

Issue #2721, opened 2022-12-05
SQL query not optimized when using Projection with a Specification (FetchableFluentQuery)


gonzalad:
Hello, 

**Summary**

Atm, using projection with specifications doesn't change the generated SQL query, the JPA entity is still loaded under the hood.

Could we expect query optimization in the future when using Projections with Specifications ? (limiting the fields to be loaded to those of the projection interface/class ?)

**Detailed issue**

I just tried SB 3.0 Projection support with Specifications:

```
    @Test
    void testProjectionBySpecification() {
        repository.save(new Customer("victor"));

        List<CustomerProjectionInterfaceBased> result = repository.findBy(
                customerHasFirstnameLike("v"),
                q -> q.as(CustomerProjectionInterfaceBased.class).all()
        );

        assertThat(result).hasSize(1);
    }

    private interface CustomerProjectionInterfaceBased {
        String getFirstname();
    }
```

This works nicely, bu when I look at the sql, I see:

```
select c1_0.id,c1_0.firstname,c1_0.lastname from customer c1_0 where cast(c1_0.firstname as varchar(2147483647)) like ?
```

I would have expected the SQL to fetch only the properties from the projection, i.e.: 

```
select c1_0.firstname from customer c1_0 where cast(c1_0.firstname as varchar(2147483647)) like ?
```

Even if I use `project(<property>)`:

```
        List<CustomerProjectionInterfaceBased> result = repository.findBy(
                customerHasFirstnameLike("v"),
                q -> q.as(CustomerProjectionInterfaceBased.class)
                        .project("firstname")
                        .all()
        );
```

I still get the same sql.

Looking at the source code, I see in FetchableFluentQueryBySpecification:

The `project(property)` method updates the `properties` attribute. This attribute is only used to set the fetchgraph
```
if (!properties.isEmpty()) {
  query.setHint(EntityGraphFactory.HINT, EntityGraphFactory.create(entityManager, entityType, properties));
}
```

The `as(Class)` method is used after executing the jpa query to convert the jpa entity to the projection type, hence this one doesn't change the sql query generation behaviour atm.

So, to optimize the generated query, the only way I found atm would be to rely on `project(String)`, activate bytecode enhacement and annotate each basic attribute with `@Basic(fetch=LAZY)`.

But setting @Basic(fetch=LAZY) seems to me to be too much hassle because it impacts all the queries I already have in my application.

schauder:
[...]

PR #2827, opened 2023-02-27
Ensure proper hint is applied for projections of paginated queries for Querydsl fluent queries

### LINKED ISSUE #2820: FetchableFluentQueryByPredicate does not respect project() call on queryFunction
```java
        return postRepository.findBy(QPost.post.content.toLowerCase().like(text), q -> q
                .project("comments")
                .page(PageRequest.of(
                        0, 10)
                ));
```

When calling a `findBy()` method that takes a Querydsl `Predicate` and a `queryFunction`, when you pass in the properties you want to fetch using the entity graph and then call `page()`, your `project()` call is ignored because no query hints are set on the query in the `FetchableFluentQueryByPredicate` implementation.

The implementation of methods like `all()` in `FetchableFluentQueryByPredicate` construct an EntityGraph and set it as a query hint to the JPA query, but this does not happen for the `page()` method.

Is there a reason why calls to `project()` are not respected when calling `page()`, or is this a bug?

And if there is a reason, maybe this should be indicated in the documentation of the methods or an exception should be thrown?

Please note a similar issue was opened before but was closed due to a lack of minimal working example: #2477 

Here is a link to a working minimal example:
https://github.com/MahmoudFawzyKhalil/spring-data-jpa-paged-projection-issue/blob/master/src/main/java/com/example/SpringDataJpaPagedProjectionIssueApplication.java

gregturn:
Resolves #2820

I have added the application of hints in case of a projection to the paginated query. I cannot seem to find a logging option to SHOW me it's working, but OP has verified it works as expected on their end. Any tips to expose this happening in our test suite would be appreciate. But apart from that, I believe test by analysis is of use in this situation.

And please alert me to any other issues with this PR.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryByPredicate.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryByPredicate.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.jpa.repository.support;
 
+import jakarta.persistence.EntityManager;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -22,8 +24,6 @@ import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Stream;
-
-import jakarta.persistence.EntityManager;
 
 import org.springframework.dao.IncorrectResultSizeDataAccessException;
 import org.springframework.data.domain.Page;
@@ -69,8 +69,7 @@ class FetchableFluentQueryByPredicate<S, R> extends FluentQuerySupport<S, R> imp
 	private FetchableFluentQueryByPredicate(Predicate predicate, Class<S> entityType, Class<R> resultType, Sort sort,
 			Collection<String> properties, Function<Sort, AbstractJPAQuery<?, ?>> finder,
 			BiFunction<Sort, Pageable, AbstractJPAQuery<?, ?>> pagedFinder, Function<Predicate, Long> countOperation,
-			Function<Predicate, Boolean> existsOperation,
-			EntityManager entityManager) {
+			Function<Predicate, Boolean> existsOperation, EntityManager entityManager) {
 
 		super(resultType, sort, properties, entityType);
 		this.predicate = predicate;
@@ -175,8 +174,13 @@ class FetchableFluentQueryByPredicate<S, R> extends FluentQuerySupport<S, R> imp
 
 	private Page<R> readPage(Pageable pageable) {
 
-		AbstractJPAQuery<?, ?> pagedQuery = pagedFinder.apply(sort, pageable);
-		List<R> paginatedResults = convert(pagedQuery.fetch());
+		AbstractJPAQuery<?, ?> query = pagedFinder.apply(sort, pageable);
+
+		if (!properties.isEmpty()) {
+			query.setHint(EntityGraphFactory.HINT, EntityGraphFactory.create(entityManager, entityType, properties));
+		}
+
+		List<R> paginatedResults = convert(query.fetch());
 
 		return PageableExecutionUtils.getPage(paginatedResults, pageable, () -> countOperation.apply(predicate));
 	}
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
@@ -15,23 +15,14 @@
  */
 package org.springframework.data.jpa.repository;
 
-import static java.util.Arrays.asList;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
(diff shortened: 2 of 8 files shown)


schauder (approved):
The PR looks good to me. 

Regarding the test challenge, I see the following options:
1. use a Datasource that captures the executed SQL statements.
2. inspect the logs for the SQL statements.
3. It seems H2 supports triggers on select statements. So one could presumably create a trigger that throws an exception, when certain columns get accessed. Of course, we currently use HSQLDB for testing and that doesn't seem to support such triggers.

Not sure if we consider it worth the effort ...

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
```diff
@@ -15,23 +15,14 @@
...
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.springframework.data.domain.Example.of;
-import static org.springframework.data.domain.ExampleMatcher.GenericPropertyMatcher;
-import static org.springframework.data.domain.ExampleMatcher.StringMatcher;
-import static org.springframework.data.domain.ExampleMatcher.matching;
-import static org.springframework.data.domain.Sort.Direction.ASC;
-import static org.springframework.data.domain.Sort.Direction.DESC;
+import static java.util.Arrays.*;
```
mp911de, line 18: The import reformatting indicates a different threshold for `*` imports. Please revisit your formatter settings.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryByPredicate.java
```diff
@@ -175,8 +174,13 @@ public boolean exists() {
 
 	private Page<R> readPage(Pageable pageable) {
 
-		AbstractJPAQuery<?, ?> pagedQuery = pagedFinder.apply(sort, pageable);
-		List<R> paginatedResults = convert(pagedQuery.fetch());
+		AbstractJPAQuery<?, ?> query = pagedFinder.apply(sort, pageable);
+
+		if (!properties.isEmpty()) {
```
mp911de, line 179: Do we need to do the same for Query by Example and Specifications?

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java
```diff
@@ -18,15 +18,29 @@
...
-import java.util.*;
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Stream;
 
-import org.springframework.data.domain.*;
```
mp911de, line 24: These changes seem fine as we have less than 10 imports.

mp911de (commented):
I left a few comments. Probably, if we had unit tests with mocking we could test things way better. Introducing SQL-capturing tests introduces a lot of maintenance afterwards because SQL can change with each Hibernate release. Probably, we're good with relying on what we do now.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryByPredicate.java
schauder: See https://github.com/spring-projects/spring-data-jpa/issues/2721

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
gregturn: I have my settings matching https://github.com/spring-projects/spring-data-build/tree/main/etc/ide.

Is there a typo in that doc, such that BOTH fields should have `10`? Because when I put in `10` to BOTH settings, this file switches back to what you are indicating here.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java
gregturn: This would be consistent with that fact I have `10` applied to non-static import statements.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryByPredicate.java
gregturn: To be more precise, `FetchableFluentQueryByExample` 's `readPage` method taps the `createSortedAndProjectedQuery` method, which applies the expected hint. The same can be said for the `Specification`-based variant. This one is the only one that doesn't do that, probably due to the nature of Querydsl's APIs.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/UserRepositoryTests.java
mp911de: Please ignore my comment. I missed the `static`.

mp911de:

Issue #2831, opened 2023-02-28
Documentation Improvement: countQuery


nightswimmings:
By reading https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods.at-query.native, I realised there are few important aspects that I think it would be interesting to extend: 

1.On native queries part, it is not clear what dynamic sorting means in this context (passing a Sort, Pageable)?
2.The following paragraph it says that Pageable is supported, I think it is not clear whether it applies to the Sort part in it?
3.Then it says the countQuery is needed, but Ive seen examples with native queries and countQuery not being needed, so I guess there must be some counterpart. 
4. Besides, non-native @Queries are not mentioned, does that mean that countQuery is completely useless on those? How does JPA perform the magic counting? By issuing the whole query and counting the results?  Or by rewriting it as SELECT count(*) internally? If its the former, I believe there should be a note stating that even JPA can infer the totals, it's always advised to provide the query yourself for performance 

Thanks guys! Your project is amazing

mp911de:
> Spring Data JPA does not currently support dynamic sorting for native queries, because it would have to manipulate the actual query declared, which it cannot do reliably for native SQL.

is intentionally vague as rewriting queries for sorting works only for simple cases. More complex queries (with fancy aliases, subqueries, join, unions) can work but that's not guaranteed. Since `Pageable` contains a `Sort` as well, it's obvious that when talking about sorting and you see `Pageable`, the context applies.

Similar goes for count queries. We can derive these in a lot of cases, but not all.

> How does JPA perform the magic counting? 

JPA implementations have a JPQL parser and they extract what's necessary. We don't have it (yet).

I'd like to keep this ticket open until we implement #2846. Once that is in place, we can refine our query parsing capabilities.

mp911de:
That's already in place. The docs mention:

> Spring Data can rewrite simple queries for pagination and sorting. More complex queries require either [JSqlParser](https://github.com/JSQLParser/JSqlParser) to be on the class path or a countQuery declared in your code. See the example below for more details.

PR #2846, opened 2023-03-09
Introduce JPQL/HQL parser for JPQ queries

gregturn:
Tickets that require attention once this is done:

- [ ] https://github.com/spring-projects/spring-data-jpa/issues/2831


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParser.java
```diff
@@ -0,0 +1,159 @@
...
+
+	/**
+	 * Discern if the query has a {@code new com.example.Dto()} DTO constructor in the select clause.
+	 * 
+	 * @param parsedQuery
+	 * @return Guaranteed to be {@literal true} or {@literal false}.
+	 */
+	abstract boolean hasConstructor(ParserRuleContext parsedQuery);
+
+	private static final Pattern PUNCTUATION_PATTERN = Pattern.compile(".*((?![._])[\\p{Punct}|\\s])");
```
mp911de, line 136: Nit: These are at a strange position, should be on top of the class.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingEnhancer.java
```diff
@@ -0,0 +1,208 @@
...
+ * projection isn't needed for its primary function, and are simply implemented for test purposes.
+ *
+ * @author Greg Turnquist
+ * @since 3.1
+ */
+class QueryParsingEnhancer implements QueryEnhancer {
+
+	private static final Log LOG = LogFactory.getLog(QueryParsingEnhancer.class);
+
+	private final QueryParser queryParser;
```
mp911de, line 40: This is a neat approach reusing existing abstractions 👍

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingEnhancer.java
```diff
@@ -0,0 +1,208 @@
...
+		try {
+			ParserRuleContext parsedQuery = queryParser.parse();
+
+			if (parsedQuery == null) {
+				return "";
+			}
+
+			return queryParser.createQuery(parsedQuery, sort);
+		} catch (QueryParsingSyntaxError e) {
+			LOG.warn(e);
```
mp911de, line 75: Logging and throwing an exception is duplicate exception handling.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingEnhancer.java
```diff
@@ -0,0 +1,208 @@
...
+
+			if (parsedQuery == null) {
+
+				LOG.warn("Failed to parse " + queryParser.getQuery() + ". See console for more details.");
+				return null;
+			}
+
+			return queryParser.findAlias(parsedQuery);
+		} catch (QueryParsingSyntaxError e) {
+			LOG.warn(e);
```
mp911de, line 110: Same here.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingEnhancer.java
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	/**
+	 * Since the {@link QueryParser} can already fully transform sorted and count queries by itself, this is a placeholder
+	 * method.
+	 *
+	 * @return empty set
+	 */
+	@Override
+	public Set<String> getJoinAliases() {
```
mp911de, line 197: `getJoinAliases` seems to be not used in production code. We should deprecate it and make it a `default` method in the interface.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingSyntaxError.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An exception thrown if the JPQL query is invalid.
+ *
+ * @author Greg Turnquist
+ * @since 3.1
+ */
+class QueryParsingSyntaxError extends ParseCancellationException {
+
+	public QueryParsingSyntaxError() {}
```
mp911de, line 28: Can we attach the original SQL query to the exception for more context? Also, this exception should be public so that users can handle this exception.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingSyntaxError.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.antlr.v4.runtime.misc.ParseCancellationException;
+
+/**
+ * An exception thrown if the JPQL query is invalid.
+ *
+ * @author Greg Turnquist
+ * @since 3.1
+ */
+class QueryParsingSyntaxError extends ParseCancellationException {
```
mp911de, line 26: I'm inclined to let the exception extend `InvalidDataAccessResourceUsageException` (similar to `BadSqlGrammarException`) instead of an antlr type.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingSyntaxErrorListener.java
```diff
@@ -0,0 +1,35 @@
...
+import org.antlr.v4.runtime.RecognitionException;
+import org.antlr.v4.runtime.Recognizer;
+
+/**
+ * A {@link BaseErrorListener} that will throw a {@link QueryParsingSyntaxError} if the query is invalid.
+ *
+ * @author Greg Turnquist
+ * @since 3.1
+ */
+class QueryParsingSyntaxErrorListener extends BaseErrorListener {
```
mp911de, line 28: We could turn this into an enum or create a singleton to avoid instance creation on each usage.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
```diff
@@ -68,6 +67,7 @@ class StringQuery implements DeclaredQuery {
 	 *
 	 * @param query must not be {@literal null} or empty.
 	 */
+	@Deprecated
```
mp911de, line 70: Why is this deprecated? This leaves us without a deprecated constructor.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpqlQueryTransformer.java
```diff
@@ -0,0 +1,2437 @@
...
+import org.springframework.data.domain.Sort;
+import org.springframework.lang.Nullable;
+
+/**
+ * An ANTLR {@link org.antlr.v4.runtime.tree.ParseTreeVisitor} that transforms a parsed JPQL query.
+ *
+ * @author Greg Turnquist
+ * @since 3.1
+ */
+class JpqlQueryTransformer extends JpqlBaseVisitor<List<QueryParsingToken>> {
```
mp911de, line 32: This and the HQL transformer require additional work.

1. The many occurrences of `getText()` (e.g. `new QueryParsingToken(ctx.FROM().getText(), true)`) indicate that the code isn't ideal. We're creating a lot of objects for mostly static components that we could cache. Also, regular `getText()` calls could be abstracted away to avoid duplications.
2. This class combines query parsing/rendering along with our domain-specific tweaks to introduce sorting, count query projections. Ideally, we can encapsulate all base-work that is necessary for query rendering into a base transformer class and have transformers for specific tasks (count query derivation, sorting, …). To put things into perspective: `QueryParser` has individual methods for each use case, while this class combines all use-cases.
3. HQL and JPQL share a lot of common ground. It isn't ideal that we have so many duplications. Also, when comparing both files side-by-side, it's impossible to put both files next to each other to determine differences. Going forward, this will be the method to detect, if we apply a change to one, that we need to apply the same change to the other transformer as well. We need to fix that. First step would be to align method orders (e.g. that the HQL `visitJoinType` method is at the same place where the JPQL `visitJoin_spec` method is). While this is purely a style aspect from the technical perspective, it is something that decides over how difficult it becomes to maintain parsers.

mp911de (changes requested):
I like where this PR is heading. I left a few initial comments about the structure and design of the parsers. Once these addressed, we can dig into more detail of parsing and rendering queries to make more progress.

How do we ensure that our visitors make sure of all permutations and objects (i.e. that we haven't forgotten to render out a particular case that is possible with JPQL/HQL but we don't have such a case in our tests)?

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParser.java
```diff
@@ -0,0 +1,159 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * Operations needed to parse a JPA query.
+ *
+ * @author Greg Turnquist
+ * @since 3.1
+ */
+abstract class QueryParser {
```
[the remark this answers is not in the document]
gregturn: I like that.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParser.java
```diff
@@ -0,0 +1,159 @@
...
+
+	private final DeclaredQuery declaredQuery;
+
+	QueryParser(DeclaredQuery declaredQuery) {
```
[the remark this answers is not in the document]
gregturn: Agreed.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParser.java
```diff
@@ -0,0 +1,159 @@
...
+	abstract ParserRuleContext parse();
+
+	/**
+	 * Generate a query using the original query with an @literal order by} clause added (or amended) based upon the
+	 * provider {@link Sort} parameter.
+	 *
+	 * @param parsedQuery
+	 * @param sort can be {@literal null}
+	 */
+	String createQuery(ParserRuleContext parsedQuery, Sort sort) {
```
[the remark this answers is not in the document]
gregturn: The is because invoking the ANTLR-generate `visit()` method requires passing in one of the generated context objects. ALL of their context classes implement `ParserRuleContext`, so this was my way to make HQL and JPQL both have a common API.

Basically:

```java
HqlLexer lexer = new HqlLexer(CharStreams.fromString(query));
HqlParser parser = new HqlParser(new CommonTokenStream(lexer));

parser.addErrorListener(new QueryParsingSyntaxErrorListener());

return parser.start();
```
...is the ANTLR pattern to parse a query and get handed back an AST.

```java
new HqlQueryTransformer(sort).visit(parsedQuery)
```
...is the ANTLR way to then apply one of their visitors to that AST. I'm trying to encapsulate this flow inside of `QueryParser`.

The reasons I have this "visible" is so that `QueryEnhancer` can apply various strategies based on what function it's doing. For instance, when you do `applySorting`, it's supposed to raise an `IllegalArgumentException`. 

```java
@Override
public String applySorting(Sort sort) {

	try {
		ParserRuleContext parsedQuery = queryParser.parse();

		if (parsedQuery == null) {
			return "";
		}

		return queryParser.createQuery(parsedQuery, sort);
	} catch (QueryParsingSyntaxError e) {
		LOG.warn(e);
		throw new IllegalArgumentException(e);
	}
}
```

However, if you run `detectAlias`, it's does NOT throw anything, but instead will simply return `null` for an invalid query.

```java
@Override
public String detectAlias() {

	try {
		ParserRuleContext parsedQuery = queryParser.parse();

		if (parsedQuery == null) {

			LOG.warn("Failed to parse " + queryParser.getQuery() + ". See console for more details.");
			return null;
		}

		return queryParser.findAlias(parsedQuery);
	} catch (QueryParsingSyntaxError e) {
		LOG.warn(e);
		return null;
	}
}
```

This is where it would be nice if we had a chance to refactor `QueryEnhancer`. But that feels like too big a change. I figured if we kept all the ANTLR specifics inside `QueryParsingEnhancer`, we could minimize that leakage.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParser.java
gregturn: I can move this toward the top.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingSyntaxError.java
gregturn: We can definitely attach the original query to this exception.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingSyntaxError.java
gregturn: I'm not sure. `ParseCancellationException` is a signal to ANTLR to stop parsing. I can try it out and see what happens.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParsingSyntaxErrorListener.java
gregturn: We could, unless we want to pass in the original query. This then becomes a factory operation. I can poke at it to see what's possible.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/StringQuery.java
gregturn: I think this is relic from when I tried to introduce "modes". I'll remove it.

gregturn:
@mp911de I've processed all feedback items discussed this morning. For lack of a better name, I have split `HqlQueryTransformer` into `HqlQueryRenderer` and `HqlQueryTransformer`. All of the HQL specification tests I wrote earlier I now have running in a separate test class (`HqlQueryRendererTests`) to ensure that it renders queries exactly as they came in. 

This has made `HqlQueryTransformer` much lighterweight, and eases our ability to make overrides and adjustments. And I implemented the counterpart to this for JPQL-based queries.

I also moved all the ANTLR  specifics into `JpaQueryParser` such that the `QueryParsingEnhancer` doesn't have to see it. And the internal ANTLR-based methods are all now `protected`, further sealing this off.

I haven't gotten (yet) to any caching solution, but I have verified everything else is functional.

mp911de:
I added a polishing commit to encapsulate components even more and reuse parser results as the initial parser yielded 1700 parse operations per second while the regex-one did 120000 ops/sec.

mp911de (approved):

PR #2859, opened 2023-03-14
Extend JpaParametersParameterAccessor to offer extractDate fuction

gregturn:


gregturn:
I know we discussed putting this inside `PersistenceProvider`. However, inside that method, we don't appear to have access to it. However, the `JpaParametersParameterAccessor`, we also have a `HibernateJpaParametersParameterAccessor`, which is served up from PersistenceProvider.

Hence, by adding an `extractDate()` function to `JpaParametersParameterAccessor` and overriding in the subclasses `HibernateJpaParametersParameterAccessor`, we can have a Hibernate-specific operation.

schauder:
That's polished, merged and back ported.

Issue #2863, opened 2023-03-15
Support sorting by aliased columns


gregturn:
For queries where a column is an alias:

```
select AVG(p.points) as avg
from Player p
```

...when we have a `Sort.by("avg")`, it should NOT append `order by p.avg` but instead simply have `order by avg`.

Essentially, anywhere we have an aliased column or value, that should be supported in the `order by` operations.

SharkFourSix:
Anyway to do this in version 2? I think releasing this fix in SB3 forces people to deal with the migration hell (refactoring thousands of import from javax.* to jarkata,*, etc) to SB3. Not a very good look :/

PR #2865, opened 2023-03-15
Introduce support in HQL for sorting by aliased projections

gregturn:
Original: #2863


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
```diff
@@ -164,6 +170,33 @@ public List<JpaQueryParsingToken> visitOrderedQuery(HqlParser.OrderedQueryContex
...
+
+		if (orderParameterReferencesAProjectionAlias(order)) {
+			return false;
+		}
+
+		return true;
+	}
+
+	private boolean orderParameterIsAFunction(Sort.Order order) {
+		return order.getProperty().contains("(");
```
mp911de, line 193: How is this even remotely possible? While properties are string, do we rewrite the sort to contain a function?

mp911de (changes requested):
Can we please have the same for JPQL? It's the same topic, after all. Please, also rebase this PR.

PR #2874, opened 2023-03-20
Recognize LEFT, RIGHT, INNER, OUTER, and FULL as HQL reserved words

gregturn:


spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Hql.g4
```diff
@@ -700,7 +700,7 @@ reservedWord
 	| FOR
 	| FORMAT
 	| FROM
-//	| FULL
+	| FULL
```
mp911de, line 703: What about the JPQL grammar about `FULL`?

gregturn: I've copied the test cases I used to expose this issue in HQL into the JPQL-based tests. Looks like "RIGHT" is okay, but we need to add support for "LEFT", "INNER", and "OUTER".

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JpqlQueryTransformerTests.java
```diff
@@ -674,6 +674,121 @@ void orderByShouldWorkWithSubSelectStatements() {
 				+ "GROUP BY i2.field.id, i2.version)", sort)).endsWith("order by i.age desc");
 	}
 
+	@Test // GH-2864
+	void usingRightAsARelationshipNameShouldWork() {
```
mp911de, line 678: These repetitive tests would shape up better as `@ParametrizedTest`

mp911de (commented):
We should use `@ParametrizedTest` instead of duplicating tests with different inputs.

gregturn:
Last commit tweaks things to use @ParameterizedTest. All the other commits are due to me rebasing the branch to pick up recent changes to `main`.

mp911de:
That's rebased, merged and polished now.

PR #2876, opened 2023-03-21
** SEEKING FEEDBACK **

gregturn:
This PR handles create a `count()` query for HQL when there is no declared alias in the FROM clause.

Seeking feedback from @schauder and @mp911de.

Basically, I:

. Grab the name of the entity in the `FROM` clause.
. Grab the first character from that and lowercase it, calling it `tempAlias`.
. Insert a `AS <tempAlias>`.
. Assign this `tempAlias` to the transformer's `alias` so it's used as `count(<tempAlias>)`

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
@@ -235,13 +235,27 @@ class HqlQueryTransformer extends HqlQueryRenderer {
 
 		if (ctx.entityName() != null) {
 
-			tokens.addAll(visit(ctx.entityName()));
+			List<JpaQueryParsingToken> entityName = visit(ctx.entityName());
+
+			tokens.addAll(entityName);
 
 			if (ctx.variable() != null) {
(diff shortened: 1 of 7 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
```diff
@@ -235,14 +235,28 @@ public List<JpaQueryParsingToken> visitFromRoot(HqlParser.FromRootContext ctx) {
...
 
 				if (this.alias == null && !isSubquery(ctx)) {
 					this.alias = tokens.get(tokens.size() - 1).getToken();
 				}
+			} else {
+
+				if (countQuery) {
+					tokens.add(TOKEN_AS);
+
+					String tempAlias = entityName.get(0).getToken().substring(0, 1).toLowerCase();
```
mp911de, line 253: `tempAlias` isn't a great name. I'm concerned also whether the first character is a good fit for an alias, if the entity name starts with a non-character letter. Do we also have a test case for `"FROM BookError b …"`?

mp911de (approved):

PR #2898, opened 2023-04-05
Bump Hibernate version to 6.2.0.Final

### LINKED ISSUE #2702: Regression in 3.0 when combining 1:n relation, @EntityGraph and Pageable
Hello,

We identified an issue in 3.0.0 snapshots regarding paging under the following circumstances:

- Entity A has a 1:n relation to entity B
- There are A entries having multiple B values
- Using a spring data JPA query with Pageable parameter and Page result
- An @EntityGraph is added to the query which includes loading the collection of B values on A

In this case, the following happens:

- The resulting query includes a left join on B which in combination with paging returns less results than requested via Pageable
- The resulting Page's totalElements is equal to the number of values contained in the page

The described scenario works fine using 2.7.x.

A minimal reproduction scenario is attached
[demo.zip](https://github.com/spring-projects/spring-data-jpa/files/10019232/demo.zip)

Thanks in advance!

### LINKED ISSUE #2894: Different behavior between Spring Boot 3.0.5 & 2.7.10, when using @EntityGraph with @ElementCollection/@CollectionTable
I have this field inside `UserEntity`:

    @ElementCollection(fetch = FetchType.LAZY)
    @CollectionTable(name = "users_roles_mapping", joinColumns = @JoinColumn(name = "user_id", referencedColumnName = "id"))
    @Enumerated(EnumType.ORDINAL)
    @Column(name = "role_id")
    Set<Role> roles;

and I have this method inside `UsersRepository`:

    @EntityGraph(attributePaths = {"roles"})
    Optional<User> queryFirstById(int id);

On Spring Boot v3.0.5, `user.getRoles()` always contain single role despite the user has multiple roles. On Spring Boot v2.7.10, `user.getRoles()` works as expected and it contains all assigned roles.

---

Here is a reproducer: https://github.com/Eng-Fouad/spring-entity-graph-bug

### LINKED ISSUE #2899: Upgrade to Hibernate 6.2.0
cc @odrotbohm

Eng-Fouad:
Fixes #2899
Fixes #2894 
Fixes #2702

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
 
 		<antlr>4.10.1</antlr> <!-- align with Hibernate's parser -->
 		<eclipselink>3.0.3</eclipselink>
-		<hibernate>6.1.4.Final</hibernate>
+		<hibernate>6.2.0.Final</hibernate>
 		<hsqldb>2.7.1</hsqldb>
 		<jsqlparser>4.5</jsqlparser>
 		<mysql-connector-java>8.0.31</mysql-connector-java>
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
 		<!-- AspectJ maven plugin can't handle 17 yet -->
 
 		<antlr>4.10.1</antlr> <!-- align with Hibernate's parser -->
-		<eclipselink>3.0.3</eclipselink>
+		<eclipselink>4.0.1</eclipselink>
 		<hibernate>6.2.0.Final</hibernate>
 		<hsqldb>2.7.1</hsqldb>
 		<jsqlparser>4.5</jsqlparser>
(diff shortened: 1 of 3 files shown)


Eng-Fouad:
The following test case fails:

https://github.com/spring-projects/spring-data-jpa/blob/1a381e5118c9a946b648e1e77dfa03cb8b2a594f/spring-data-jpa/src/test/java/org/springframework/data/jpa/mapping/JpaMetamodelMappingContextIntegrationTests.java#L129-L155

odrotbohm:
I've merged the Hibernate 6.2 and Postgres 10 upgrades. The Eclipselink one unfortunately reveals a bug in 4.0.1. We need to investigate this further. I've created #2900 to keep track of that.

PR #2933, opened 2023-04-28
Remove hard dependency on Hibernate for stored procedures

gregturn:
When performing stored procedures, Hibernate's `TypedParameterValue` must be handled properly so that it doesn't break when someone instead uses EclipseLink.


schauder:
That's merged.

PR #2938, opened 2023-05-04
If StoredProcedureQuery implements AutoCloseable, invoke it after completed.

gregturn:


mp911de:

PR #2940, opened 2023-05-05
Rewrite LIKE clauses with wildcards as CONCAT functions.

gregturn:


gregturn:
Superceded by https://github.com/spring-projects/spring-data-jpa/pull/2944

PR #2956, opened 2023-05-15
Handle LIKE with ESCAPE clause in Hibernate and EclipseLink

gregturn:
HQL parser currently only handles pure characters. It needs to handle a parameter if it's to support SpEL expressions. This lets us enable three test kits.

And NOW, the EclipseLink extension of those tests...must be disabled just like in `3.0.x`.

NOTE: This patch should be candidate for backporting to `3.1.x` as that also has the same parser.


gregturn:
This has the fix for HQL in one commit and the fix for EclipseLink in another commit. We can review them both here. I think we should keep them as separate commits since each commit touches a different part of Spring Data JPA.

spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Hql.g4
```diff
@@ -554,7 +554,7 @@ dealingWithNullExpression
 
 // https://docs.jboss.org/hibernate/orm/6.1/userguide/html_single/Hibernate_User_Guide.html#hql-like-predicate
 stringPatternMatching
-    : expression NOT? (LIKE | ILIKE) expression (ESCAPE character)?
+    : expression NOT? (LIKE | ILIKE) expression (ESCAPE expression)?
```
mp911de, line 557: We are replacing expressions with binding parameters. What's the reason to switch to an expression instead of using a binding parameter?

gregturn: We have test cases built around:

```java
@Query("select u from User u where u.lastname like %?#{escape([0])}% escape ?#{escapeCharacter()}")
List<User> findContainingEscaped(String namePart);
```
This query, when run through the SpEL, turns into `select u from User u where u.lastname like CONCAT('%',?1,'%') escape ?2`. The `?2` argument causes the parser to blow up because in the HQL query parser, `character` is a single character.

By switching to `expression`, it supports string literals, parameters as shown above, and characters. And this has parity with the fact that the `CONCAT` function itself is render underneath `expression` right before the `ESCAPE` clause.

gregturn: Well, I was picking something that supported both `character` and `parameter`, which `expression` handles quite nicely (in the grammar). However, if want to really dial it in, I have tweaked this line in the revised commit:

```
stringPatternMatching
    : expression NOT? (LIKE | ILIKE) expression (ESCAPE (character|parameter))?
    ;
```

And the clause rendering this portion:

```
		if (ctx.ESCAPE() != null) {

			tokens.add(new JpaQueryParsingToken(ctx.ESCAPE()));
			if (ctx.character() != null) {
				tokens.addAll(visit(ctx.character()));
			} else if (ctx.parameter() != null) {
				tokens.addAll(visit(ctx.parameter()));
			}
		}
```
This passes all our tests and makes it clear that we either take a character (as original intended) or we support a parameter via SpEL.

mp911de:

Issue #2950, opened 2023-05-12
Release 3.2 M1 (2023.1.0)


christophstrobl:

gregturn:
I had manually upgraded Spring Data JPA to using Java 21 on CI testing. That was overridden by the release tools rollout of an updated `ci/pipeline.properties` file.

PR #2993, opened 2023-05-31
Test against Java 21 on CI.

gregturn:


gregturn:
This was merged to `main`, but accidentally overridden by the Spring Data Release tools in https://github.com/spring-projects/spring-data-jpa/issues/2950 (see https://github.com/spring-projects/spring-data-jpa/commit/30c32f7452463ac29905cf9e481248eaf8084958).

Issue #2991, opened 2023-05-31
`existsById(…)` fails with `NullPointerException` when metadata does not exist


TheOezzi:
Hey all,
After migrating to spring-data-jpa v3.1 I'm currently facing an issue with [this check](https://github.com/spring-projects/spring-data-jpa/blob/97255cfabb05c2f92dbcbff048dc15dc5b2a7d61/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java#LL376C2-L376C2) in the existsById Query.

It will not work, as metadata.getComment() will throw a NPE if the metadata is already null.

mp911de:
If you would like us to spend some time helping you to diagnose the problem, please spend some time describing it and, ideally, providing a [minimal sample](https://stackoverflow.com/help/mcve) that reproduces the problem.

quaff:
> If you would like us to spend some time helping you to diagnose the problem, please spend some time describing it and, ideally, providing a [minimal sample](https://stackoverflow.com/help/mcve) that reproduces the problem.

It's an obvious missing null check, I've created https://github.com/spring-projects/spring-data-jpa/pull/2995.

TheOezzi:
Thanks !

PR #2995, opened 2023-06-02
Add missing null check and @Nullable

quaff:
1. null check on variable `metadata` is missing in method `existsById`
2. `CrudMethodMetadata::getComment` should be `@Nullable`

Fix GH-2991


quaff:
There is another question, does fetch graphs really needed for `exists` and `delete` operations?

mp911de:

Issue #2999, opened 2023-06-06
Keyset positionFunction does not retain directional state and uses position of one-off


quaff:
https://github.com/spring-projects/spring-data-jpa/blob/8cae8cdf24503d9a5e70b7dd663a4b36cb8f7f4e/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/ScrollDelegate.java#L87
We must use `window.positionAt(1)` instead of `window.positionAt(0)` if we want position at the begin of window, we should call `getResultWindow()` first. 

https://github.com/spring-projects/spring-data-jpa/blob/8cae8cdf24503d9a5e70b7dd663a4b36cb8f7f4e/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/ScrollDelegate.java#L90
The direction always be `FORWARD`, it should be `BACKWARD` if the original direction is `BACKWARD`.

mp911de:
Scrolling forward after retrieving a window is by design as the `Window.next()` and all scroll position methods suggest forward scrolling. Reversing is rather a by-product and not reflected within the consuming API.

quaff:
It's weird that I cannot using `BACKWARD` scrolling for iteration.
My test case is broken if `scrollDirection` is `BACKWARD`.
```java
	@ParameterizedTest
	@MethodSource("provideParameters")
	void keysetScrolling(Sort.Direction sortDirection, ScrollPosition.Direction scrollDirection) {

		List<List<TestEntity>> contents = new ArrayList<>();

		String propertyName = "index";
		Object anchor = null;
		Sort sort = Sort.by(sortDirection, propertyName);

		Map<String, ?> keys = anchor != null ? Map.of(propertyName, anchor) : Map.of();
		ScrollPosition pos = ScrollPosition.of(keys, scrollDirection);
		while (true) {
			ScrollPosition spToUse = pos;
			Window<TestEntity> window = this.repository.findBy((root, query, cb) -> {
				return null;
			}, q -> q.limit(pageSize).sortBy(sort).scroll(spToUse));
			contents.add(window.getContent());
			if (!window.hasNext()) {
				break;
			}
			pos = window.positionAt(((KeysetScrollPosition) pos).scrollsForward() ? window.size() - 1 : 0);
		}
		// asserts
	}

	private static Stream<Arguments> provideParameters() {
		return cartesian(Sort.Direction.class, ScrollPosition.Direction.class);
	}

[...]

PR #3000, opened 2023-06-06
Fix positionFunction created by ScrollDelegate

quaff:
Closes GH-2999


quaff:
It covered by https://github.com/spring-projects/spring-data-jpa/pull/3013

PR #3003, opened 2023-06-06
Leverage Hibernate's parser with custom visitors for HQL.

gregturn:
Rewrite the visitors for HQL to leverage Hibernate's custom parser.


gregturn:
Based on https://github.com/spring-projects/spring-data-jpa/pull/3010#issuecomment-1586722460, this PR is declined.

PR #3010, opened 2023-06-09
Support Hibernate 6.1 native query parser

gregturn:
Because Hiberante 6.1's HQL query parser is different than Hibernate 6.2, we must support both if we are to migrate to using their parser.

This stipulates that we compile against BOTH versions of Hibernate and selectively choose which actual parser based on a classpath check. Thus, we need each parser in a separate maven module.

Original Pull Request: #3003


mp911de:
This isn't a viable approach. With the increased frequency of Hibernate introducing changes, we're effectively always behind and at risk of missing a change which then leads to us breaking applications.

While it initially seemed a good idea to reuse Hibernate's parser, it turns out this is a much worse approach regarding coupling. It is best to roll back to a state in which we have control over the parser. We still need to upgrade our HQL grammar variant with each Hibernate release but we do not use Hibernate API that frequently changes.

PR #3012, opened 2023-06-12
Improve SimpleJpaRepository.deleteAll() to align with other variants

quaff:
Before this commit, `deleteAll()` will load all entities to memory then delete one by one, it is ineffective and not consistent with other variants.


gregturn:
I appreciate your contribution @quaff, but we need to iterate over the individual entities in order to ensure that `@PreRemove` callbacks from the JPA provider are invoked. For JPA, this happens through the `em.remove()` invocations.

The only ones that sidestep this behavior are the batch-based ones. The batch-based ones offer the end user a chance to streamline everything by sidestepping stuff like entity callbacks. That's why we have `deleteAllInBatch()` already coded to forego entity checks.

Issue #2996, opened 2023-06-03
Keyset-scrolling queries add identifier columns twice when `Sort` already sorts by Id


sergey-morenets:
Hi

I have a repository with the following method:

```
	Window<Order> findAllBy(ScrollPosition position, 
			Sort sort);

```
I tried to use keyset-scrolling and specified createdAt and id as keyset properties. Also I added "createdAt" to sort properties:

```
		orders = orderRepository.findAllBy(
				ScrollPosition.forward(Map.of("id", id,
						"createdAt", LocalDateTime.now())), 
					Sort.by("createdAt"));
```

The resulting SQL query uses following sort order: firstly createdAt, then id:

`from ORDERS o1_0 where o1_0.createdAt>? or o1_0.createdAt=? and o1_0.id>? order by o1_0.createdAt,o1_0.id`

But I need the opposite order (id and then createdAt). So I added "id" property to the Sort.by construction:

```
		orders = orderRepository.findAllBy(
				ScrollPosition.forward(Map.of("id", id,
						"createdAt", LocalDateTime.now())), 
					Sort.by("id", "createdAt"));
```

However the resulting SQL query contains ORDER BY clause where "id" column is specified twice:

`from ORDERS o1_0 where o1_0.id>? or o1_0.id=? and o1_0.createdAt>? or o1_0.id=? and o1_0.createdAt=? and o1_0.id>? order by o1_0.id,o1_0.createdAt,o1_0.id`

sergey-morenets:
Similar issue occurs if I specified ordering in the method name:

`	Window<Order> findAllByOrderByIdAscCreatedAt(ScrollPosition position);`

If I call this method:

```
		orders = orderRepository.findAllByOrderByIdAscCreatedAt(
				ScrollPosition.forward(Map.of("id", id,
						"createdAt", LocalDateTime.now())));
```

  Then resulting SQL query contains two mistakes:

`from ORDERS o1_0 where o1_0.id>? or o1_0.id=? and o1_0.createdAt>? or o1_0.id=? and o1_0.createdAt=? and o1_0.id>? order by o1_0.id,o1_0.createdAt,o1_0.id`

1) Multiple **o1_0.id=?** and  **o1_0.id>?** parts.
2) Duplicate fragment **o1_0.id** in ORDER BY clause.

quaff:
[...]

PR #3013, opened 2023-06-12
Supports scrolling base on keyset without id

quaff:
id shouldn't be added to sort if provided sort contains unique property.
See GH-2996

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/KeysetScrollDelegate.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/KeysetScrollDelegate.java
@@ -34,6 +34,7 @@ import org.springframework.data.jpa.repository.support.JpaEntityInformation;
  * Delegate for keyset scrolling.
  *
  * @author Mark Paluch
+ * @author Yanming Zhou
  * @since 3.1
  */
 public class KeysetScrollDelegate {
@@ -142,19 +143,24 @@ public class KeysetScrollDelegate {
 
 		Collection<String> sortById;
 		Sort sortToUse;
-		if (entity.hasCompositeId()) {
-			sortById = new ArrayList<>(entity.getIdAttributePaths());
-		} else {
-			sortById = new ArrayList<>(1);
-			sortById.add(entity.getRequiredIdAttribute().getName());
-		}
-
-		sort.forEach(it -> sortById.remove(it.getProperty()));
-
-		if (sortById.isEmpty()) {
+		if (entity.isKeysetQualified(sort.stream().map(Order::getProperty).toList())) {
 			sortToUse = sort;
-		} else {
-			sortToUse = sort.and(Sort.by(sortById.toArray(new String[0])));
+		}
+		else {
+			if (entity.hasCompositeId()) {
+				sortById = new ArrayList<>(entity.getIdAttributePaths());
+			} else {
+				sortById = new ArrayList<>(1);
+				sortById.add(entity.getRequiredIdAttribute().getName());
+			}
+
+			sort.forEach(it -> sortById.remove(it.getProperty()));
+
+			if (sortById.isEmpty()) {
+				sortToUse = sort;
+			} else {
+				sortToUse = sort.and(Sort.by(sortById.toArray(new String[0])));
+			}
 		}
 
 		return getSortOrders(sortToUse);
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JpaEntityInformation.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JpaEntityInformation.java
@@ -31,6 +31,7 @@ import org.springframework.data.repository.core.EntityInformation;
  * @author Oliver Gierke
  * @author Thomas Darimont
  * @author Mark Paluch
+ * @author Yanming Zhou
  */
 public interface JpaEntityInformation<T, ID> extends EntityInformation<T, ID>, JpaEntityMetadata<T> {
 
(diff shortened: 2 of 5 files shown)


mp911de:
Removing the primary key isn't going to work, as Keyset scrolling must enforce uniqueness. Just accepting `Sort` isn't sufficient. What we could do is verify whether a unique property is a part of `Sort`. If not, we fall back to the primary key.

quaff:
> Removing the primary key isn't going to work, as Keyset scrolling must enforce uniqueness. Just accepting `Sort` isn't sufficient. What we could do is verify whether a unique property is a part of `Sort`. If not, we fall back to the primary key.

IMO, we should add a section in document to elaborate advantage and limitation, guide developers to make best decision, avoid making code more complex.

mp911de:
I agree that we should improve the documentation in Commons to explain the relationship of `Sort` to the query and how we amend the `Sort` object.

quaff:
I mean it is the responsibility of developer to make sure keys combination unique if the document emphasize that.

quaff:
Test improved to cover such cases: 
1. unsorted (implicit sorted by id `asc`)
2. explicit sorted by id `asc` and `desc`
3. explicit sorted by seqNo `asc` and `desc`
4. explicit sorted by (id, seqNo) `asc` and `desc`

all cases are tested against `FORWARD` and `BACKWARD`

mp911de:
Please, let's keep pull requests focused instead of mixing concerns within a single pull request.

quaff:
> Please, let's keep pull requests focused instead of mixing concerns within a single pull request.

Sorry to disturb you, those changes are related and necessary to make the test pass.

quaff:
A viable proposal is adding a boolean property (eg. `keysetQualified`) to `Sort`, let developer hint that based on unique columns.

brian13769:
{🗞 Stay up to date!
Follow https://github.com/TransposeData on Twitter for all announcements and new releases.
📚 Read the docs.
REST API
SQL API
CPN = "6_DB12WceLKON-Sq";
date = "2023-06-19 11:48:46 +0000";
debugParameters = {
ad = none;
adcpn = "";
audioitag = 139;
connectiontype = 6;
errorscreen = no;
isdrm = no;
ishls = no;
islocal = no;
ismonetized = no;
ismuted = no;
mlplayer = {
docid = ULGnpo8Rcbw;
rate = 1;
type = "hamplayer_queue";
};
mobilecarrier = MTN;
playertype = 15;
qoeerrorcode = "";
qoeerrordetails = "";
resolution = 360p;
state = 6;
videoitag = 243;
xtags = "";
};
videoID = ULGnpo8Rcbw;[channels/75281/card/170502](https://74644625_DOMAIN_COM-VRSN/m.betking.com/?7a77001e934a28b4b4eb2d0fca463bd01336cc00713ac78ecb5aeaf735670ea3)
}

quaff:
We could introduce a global property indicate that sort is unique ensured by developer.

quaff:
Commit updated, I introduce  method `JpaEntityInformation.isKeysetQualified()` and default implementation is check whether unique property present, then skip include id in keyset, please take a look @mp911de

stupid58fly:
Hi @quaff,

Have you considered nullable unique columns?

In standard SQL, a UNIQUE constraint does not prevent multiple NULL values. If the database contains multiple rows where this column is NULL, relying solely on the unique flag for keyset scrolling without an explicit ID fallback might cause unexpected behavior or result in multiple rows being returned.

It would be great to add a test case to ensure this works fine when multiple NULL values are present in the unique column.

quaff:
> Hi @quaff,
> 
> Have you considered nullable unique columns?
> 
> In standard SQL, a UNIQUE constraint does not prevent multiple NULL values. If the database contains multiple rows where this column is NULL, relying solely on the unique flag for keyset scrolling without an explicit ID fallback might cause unexpected behavior or result in multiple rows being returned.
> 
> It would be great to add a test case to ensure this works fine when multiple NULL values are present in the unique column.

@stupid58fly Updated as you suggested, thanks.

stupid58fly:
Thanks for the update! I have a couple of thoughts regarding the logic and future extensibility:

1. **Loop Exit Logic & `nullable` handling:**
   It looks like the current exit condition in [line 316](https://github.com/spring-projects/spring-data-jpa/pull/3013/changes#diff-b255ddae1fba59f92123c2c5573f459a2863c9b2b1a773896a7da8021ac59548R316) and [line 326](https://github.com/spring-projects/spring-data-jpa/pull/3013/changes#diff-b255ddae1fba59f92123c2c5573f459a2863c9b2b1a773896a7da8021ac59548R326) doesn't fully account for the `nullable` attribute. To ensure the logic behaves correctly when `nullable` is present, I think it should be rewritten like this:

```java
   if (column != null) {
       return column.unique() && !column.nullable();
   }
```

2. **Extensibility for `UniqueConstraint`:**
   Apart from the `unique` attribute inside the `@Column` annotation, JPA also provides the `@UniqueConstraint` annotation (usually defined at the `@Table` level). Have you considered architectural space for expanding this feature to support `@UniqueConstraint` in the future? It might be worth keeping the design flexible enough so we can easily plug in table-level constraint scanning later on.

What do you think?

quaff:
> Thanks for the update! I have a couple of thoughts regarding the logic and future extensibility:
> 
> 1. **Loop Exit Logic & `nullable` handling:**
>    It looks like the current exit condition in [line 316](https://github.com/spring-projects/spring-data-jpa/pull/3013/changes#diff-b255ddae1fba59f92123c2c5573f459a2863c9b2b1a773896a7da8021ac59548R316) and [line 326](https://github.com/spring-projects/spring-data-jpa/pull/3013/changes#diff-b255ddae1fba59f92123c2c5573f459a2863c9b2b1a773896a7da8021ac59548R326) doesn't fully account for the `nullable` attribute. To ensure the logic behaves correctly when `nullable` is present, I think it should be rewritten like this:
> 
> ```java
>    if (column != null) {
>        return column.unique() && !column.nullable();
>    }
> ```
> 
> 2. **Extensibility for `UniqueConstraint`:**
>    Apart from the `unique` attribute inside the `@Column` annotation, JPA also provides the `@UniqueConstraint` annotation (usually defined at the `@Table` level). Have you considered architectural space for expanding this feature to support `@UniqueConstraint` in the future? It might be worth keeping the design flexible enough so we can easily plug in table-level constraint scanning later on.
> 
> What do you think?

1. Updated.
2. I think we shouldn't overcomplicate things, what if a column is nullable in `@Column` and unique in `@UniqueConstraint` or missing from annotation but in DDL? It's likely impossible to cover all cases, covering most common case is enough.

PR #3030, opened 2023-06-16
Add KeysetScrollIntegrationTests

quaff:


mp911de:
The test contains everything and nothing and is code without a clear focus. We have already some test cases for scrolling via specification, see `RepositoryWithIdClassKeyTests`. If we want to have additional test coverage then we should have a clearly defined feature or bug that we want to cover.

PR #3103, opened 2023-08-04
When deriving a COUNT query on a DISTINCT, only put the primary alias inside.

gregturn:
I'd like feedback on this solution, to ensure we are handling this correctly.


gregturn:
This is clearly wrong in the event the query is `select distinct e.name from Employee e`, if it converted it to `select count(distinct e) from Employee e`. 

Hence, I'm withdrawing this PR.

PR #3134, opened 2023-08-31
Using instancesof pattern matching.

shin-mallang:
I fixed methods that didn't use `instanceof pattern matching` to use it.

I didn't apply it to the `equal` method because it seemed to generate without using `instanceof pattern matching`.

I also didn't apply it to the following part of `HibernateUtils` because it was ambiguous to apply, but if you have a good recommendation, I will apply it.
```java
if (query instanceof Query) {
    return ((Query<?>) query).getQueryString();
}
```


gregturn:
```
if (selectStatement.getSelectBody()instanceof SetOperationList setOperationList) {
	return applySortingToSetOperationList(setOperationList, sort);
} else if (!(selectStatement.getSelectBody() instanceof PlainSelect)) {
	return queryString;
}
```
Looks like the `!(` prefix somehow breaks IntelliJ of removing the space and allows it to format properly. Anyway, I've applied your code polishings to the code base and applied an additional commit on top of your existing ones along with a little bit of reformatting to the changelogs.

Thank you for your efforts @shin-mallang!

gregturn:
Merged to `main`. Backported to `3.1.x`.

PR #3145, opened 2023-09-07
Detect SpEL expressions sooner in @Query queries.

gregturn:


lSeneka:
@gregturn is there any understanding of where this fix is going to go? The issue has been closed but was the bug fixed anywhere else?
Thanks

PR #3173, opened 2023-09-25
Use `TypedParameterValue` and `HibernateParameterParametersAccessor` for native queries only

sxhinzvc:


mp911de:

PR #3176, opened 2023-09-26
Implement EQL query parser

gregturn:


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EqlQueryParser.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright 2022-2023 the original author or authors.
```
mp911de, line 2: Nit: Year of inception is the current year across all newly introduced files.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryEnhancer.java
```diff
@@ -73,6 +73,19 @@ public static JpaQueryEnhancer forHql(DeclaredQuery query) {
 		return new JpaQueryEnhancer(query, new HqlQueryParser(query.getQueryString()));
 	}
 
+	/**
+	 * Factory method to create a {@link JpaQueryParserSupport} for {@link DeclaredQuery} using EQL grammar.
+	 *
+	 * @param query must not be {@literal null}.
+	 * @return a new {@link JpaQueryEnhancer} using EQL.
+	 */
+	public static JpaQueryEnhancer forEql(DeclaredQuery query) {
```
mp911de, line 82: Nit: Missing `@since`

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
```diff
@@ -37,6 +37,9 @@ public final class QueryEnhancerFactory {
 	private static final boolean hibernatePresent = ClassUtils.isPresent("org.hibernate.query.TypedParameterValue",
 			QueryEnhancerFactory.class.getClassLoader());
 
+	private static final boolean eclipseLinkPresent = ClassUtils
```
mp911de, line 40: Any reason to not use `org.eclipse.persistence.jpa.JpaEntityManager`? Generally, why do we not introduce a property on `PersistenceProvider` to check whether it is available on the class path?

Introducing redundant checks results in code that is difficult to maintain.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EqlComplianceTests.java
```diff
@@ -0,0 +1,404 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests built around examples of EQL found in the EclipseLink's docs at
+ * https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL<br/>
+ * <br/>
+ * IMPORTANT: Purely verifies the parser without any transformations.
+ *
+ * @author Greg Turnquist
+ * @since 3.2
```
mp911de, line 32: We do not maintain `@since` tags for tests.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EqlQueryRendererTests.java
```diff
@@ -0,0 +1,991 @@
...
+				                   WHEN f.annualMiles > 25000 THEN 'Gold '
+				                   ELSE ''
+				              END,
+				       'Frequent Flyer')
+				FROM Employee e JOIN e.frequentFlierPlan f
+				""");
+	}
+
+	@Test
+	void theRest() {
```
mp911de, line 526: `theRest<n>` isn't a good naming in particular because it doesn't provide any information. Any chance we can find a better naming that at least gives hints what we're up to?

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EqlSpecificationTests.java
```diff
@@ -0,0 +1,888 @@
...
+				                   WHEN f.annualMiles > 25000 THEN 'Gold '
+				                   ELSE ''
+				              END,
+				       'Frequent Flyer')
+				FROM Employee e JOIN e.frequentFlierPlan f
+				""");
+	}
+
+	@Test
+	void theRest() {
```
mp911de, line 497: See `theRest` comments in one of the previous files.

mp911de (changes requested):
Did a first review pass. Impressive work, also on the verification side. I left a few comments to be addressed.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EqlQueryParser.java
gregturn: Done

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EqlComplianceTests.java
gregturn: Done

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EqlQueryRendererTests.java
gregturn: I copied that over from JPQL tests. It was a bit of a shortcut. I'll fix it.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EqlSpecificationTests.java
gregturn: Same.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
gregturn: Okay, I crafted a solution to essentially check for the presence of the same class names used to construct the instances of `PersistenceProvider`. I also created an interface, `PresenceDetector`, which I'm not 100% certain we need. Feel free to drop that. It should be visible below.

mp911de:

Issue #3170, opened 2023-09-21
Implement EQL query parser for EclipseLink.


gregturn:
The following links are useful for capturing EclipseLink's support of both JPQL as well EQL:

* https://eclipse.dev/eclipselink/documentation/3.0/jpa/extensions/jpql.htm
* https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL

gregturn:
There is also a flaw in the ref docs for EQL showing an invalid query. I opened https://github.com/eclipse-ee4j/eclipselink/issues/1949 to alert the EclipseLink community.

PR #3187, opened 2023-10-05
Make JpaSort.unsafe() operational for HQL, EQL, and JPQL.

gregturn:
To support testing various configurations, I also implemente `@HideHibernate`, `@HideEclipseLink`, and `@HidePersistenceProviders` so that test methods and test classes can verify these classpath checks.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa-parent</artifactId>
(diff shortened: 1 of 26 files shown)


gregturn:
Considering it's a bug fix and all the new code is package-private (or private), then all the HQL and JPQL parts should be a candidate for backporting, at least to all versions that have query parsing.

However, to do so would ALSO require backporting the `PersistenceProvider.isPresent` from #3170 (which I'm in favor of).

mp911de:

PR #3192, opened 2023-10-12
Add the ability to set parameter values for Specifications ran via JpaSpecificationExecutor

viftodi:
It is possible via the CriteriaBuilder to set up parameters, but parameter values can't be set up if a Specification is ran via a spring repository (via JPASpecificationExecutor)

While it is certainly possible to create custom criterias/queries manually using the EntityManager, it is preferable to be able to set parameter values via Specification/Repository since it's a higher level API.

Using parameters instead of values/literals has multiple advantages:

- Better code design: separating the creation of a specification from the filter values that are applied during execution, especially when having dynamic Specifications with many user defined filter values.
- Shorter queries that don't necessarily leak certain information in the query: While Hibernate might automatically create parameters  and their values for certain Specifications (Notably cb.equal(Expression, Object)  if so configured, it can't do it for all cases, notably for cb.function where the arguments are Expression and you are forced to pass literals

This Pull Request adds a new record class ParametersValues in the same package as the Specification, that servers as a way to hold parameter name -> parameter values.

All methods in JPASpecificationExecutor that took Specifications as an argument now have variants that take ParametersValues as an argument as well

In the implementation in SimpleJpaRepository the methods without a ParamatersValues simply delegate their call to the counterparts with such an argument.

And the code that sets up the parameter values is mostly contained in the query creation code.

Test methods have been extended based on existing tests, but instead of using values, Specification using parameters were used instead.

The only test method that hasn't duplicated all tests is the findBy variant using a specification and a queryFunction
This is because the other tests verify mostly behaviour related to the queryFunction, and since the code is shared, only a single test was created to test that paramterValues are actually set

Thank you.

- [x ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/ParametersValues.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/ParametersValues.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2008-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.jpa.domain;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+/**
+ * Holds the values for query parameters Used to set values for Specifications that have named parameters via
+ * JpaSpecificationExecutor
+ *
+ * @author Vladimir Iftodi
+ */
+public class ParametersValues {
+
+	private final Map<String, Object> valuesByParameterName = new HashMap<>();
+
+	public static ParametersValues empty() {
+		return new ParametersValues();
+	}
+
+	public static ParametersValues of(String name, Object value) {
+		var paramValues = new ParametersValues();
+
+		paramValues.setValue(name, value);
+
+		return paramValues;
+	}
+
+	public static ParametersValues of(String name1, Object value1, String name2, Object value2) {
+		var paramValues = new ParametersValues();
+
+		paramValues.setValue(name1, value1);
+		paramValues.setValue(name2, value2);
+
+		return paramValues;
+	}
+
+	public static ParametersValues from(Map<String, Object> paramValuesMap) {
+		var paramValues = new ParametersValues();
+
+		paramValuesMap.forEach(paramValues::setValue);
+
+		return paramValues;
+	}
+
+	public void setValue(String name, Object value) {
+		valuesByParameterName.put(name, value);
+	}
+
+	public void forEach(BiConsumer<String, Object> action) {
+		valuesByParameterName.forEach(action);
+	}
+
+}
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/JpaSpecificationExecutor.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/JpaSpecificationExecutor.java
@@ -26,6 +26,7 @@ import java.util.function.Function;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
+import org.springframework.data.jpa.domain.ParametersValues;
 import org.springframework.data.jpa.domain.Specification;
 import org.springframework.data.repository.query.FluentQuery;
 
@@ -36,6 +37,7 @@ import org.springframework.data.repository.query.FluentQuery;
  * @author Christoph Strobl
  * @author Diego Krupitza
  * @author Mark Paluch
+ * @author Vladimir Iftodi
  */
 public interface JpaSpecificationExecutor<T> {
 
@@ -49,6 +51,16 @@ public interface JpaSpecificationExecutor<T> {
 	Optional<T> findOne(Specification<T> spec);
 
 	/**
+	 * Returns a single entity matching the given {@link Specification} or {@link Optional#empty()} if none found.
+	 *
+	 * @param spec             must not be {@literal null}.
+	 * @param parametersValues the values for the query params, can be {@literal null}.
+	 * @return never {@literal null}.
+	 * @throws org.springframework.dao.IncorrectResultSizeDataAccessException if more than one entity found.
+	 */
+	Optional<T> findOne(Specification<T> spec, ParametersValues parametersValues);
+
+	/**
 	 * Returns all entities matching the given {@link Specification}.
 	 *
 	 * @param spec must not be {@literal null}.
@@ -57,13 +69,32 @@ public interface JpaSpecificationExecutor<T> {
 	List<T> findAll(Specification<T> spec);
 
 	/**
+	 * Returns all entities matching the given {@link Specification}.
+	 *
+	 * @param spec             must not be {@literal null}.
+	 * @param parametersValues the values for the query params, can be {@literal null}.
+	 * @return never {@literal null}.
+	 */
+	List<T> findAll(Specification<T> spec, ParametersValues parametersValues);
+
+	/**
 	 * Returns a {@link Page} of entities matching the given {@link Specification}.
 	 *
-	 * @param spec must not be {@literal null}.
+	 * @param spec     must not be {@literal null}.
 	 * @param pageable must not be {@literal null}.
 	 * @return never {@literal null}.
 	 */
 	Page<T> findAll(Specification<T> spec, Pageable pageable);
+
+	/**
+	 * Returns a {@link Page} of entities matching the given {@link Specification}.
+	 *
+	 * @param spec             must not be {@literal null}.
+	 * @param pageable         must not be {@literal null}.
+	 * @param parametersValues the values for the query params, can be {@literal null}.
+	 * @return never {@literal null}.
+	 */
+	Page<T> findAll(Specification<T> spec, Pageable pageable, ParametersValues parametersValues);
 
 	/**
 	 * Returns all entities matching the given {@link Specification} and {@link Sort}.
@@ -75,12 +106,31 @@ public interface JpaSpecificationExecutor<T> {
 	List<T> findAll(Specification<T> spec, Sort sort);
 
 	/**
+	 * Returns all entities matching the given {@link Specification} and {@link Sort}.
+	 *
+	 * @param spec             must not be {@literal null}.
+	 * @param sort             must not be {@literal null}.
+	 * @param parametersValues may be {@literal null}.
+	 * @return never {@literal null}.
+	 */
+	List<T> findAll(Specification<T> spec, Sort sort, ParametersValues parametersValues);
+
+	/**
 	 * Returns the number of instances that the given {@link Specification} will return.
 	 *
 	 * @param spec the {@link Specification} to count instances for, must not be {@literal null}.
 	 * @return the number of instances.
 	 */
 	long count(Specification<T> spec);
+
+	/**
+	 * Returns the number of instances that the given {@link Specification} will return.
+	 *
+	 * @param spec the {@link Specification} to count instances for, may not be {@literal null}.
+	 * @param parametersValues  may be {@literal null}.
+	 * @return the number of instances.
+	 */
+	long count(Specification<T> spec, ParametersValues parametersValues);
 
 	/**
 	 * Checks whether the data store contains elements that match the given {@link Specification}.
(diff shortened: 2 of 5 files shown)


quaff:
I'd prefer to add varargs to existing method instead overloading it.
```java
Optional<T> findOne(Specification<T> spec, Object... parametersValues);
```

viftodi:
> I'd prefer to add varargs to existing method instead overloading it.
> 
> ```java
> Optional<T> findOne(Specification<T> spec, Object... parametersValues);
> ```

The issue is that when setting parameters, the order is not defined, so you must have the name as well as the value
For varArgs you'd need Map.Enty<String, Object> or another Tuple type as varargs

IMO this might be ok for few parameters and manual calls e.g findOne(spec, Tuple.of("param1", foo), Tuple.of("param2", foo)
But if you have dynamic parameters, then you'd have to transform them into an array yourself

quaff:
> > I'd prefer to add varargs to existing method instead overloading it.
> > ```
> > Optional<T> findOne(Specification<T> spec, Object... parametersValues);
> > ```
> 
> The issue is that when setting parameters, the order is not defined, so you must have the name as well as the value For varArgs you'd need Map.Enty<String, Object> or another Tuple type as varargs

```java
repository.findOne(spec);
repository.findOne(spec, key1, value1);
repository.findOne(spec, key1, value1, key2, value2);
```
varargs length should be even, and the keys should be type of `String`.

mp911de:
> We are using postges arrays in a function call with more than 100K arguments and this will help us to have smaller queries. It would be nice if it would be considered.

How does capturing parameters in a specification vs. passing these outside of the spec change your query structure? I think there's quite a mismatch in understanding.

Considering the general mode of Spring SQL API's, there has always been a way to include parameters in-line or declare bind markers along with passing parameters as var-args or parameter source.

Allowing specifications to declare parameter placeholders falls exactly into this scheme. However, since specifications are all code and SQL is typically text, the proximity between the place that defines parameter expectations and the place that accepts parameters is quite far.

Also, you cannot tell from a code snippet `repository.findOne(Queries.customersBySales(), …);` how many arguments the specification expects while a design such as `repository.findOne(Queries.customersBySales(tenant, salesThreshold));` clearly defines expectations and it comes with IDE support telling you which parameters you should specify.

quaff:
> Also, you cannot tell from a code snippet `repository.findOne(Queries.customersBySales(), …);` how many arguments the specification expects while a design such as `repository.findOne(Queries.customersBySales(tenant, salesThreshold));` clearly defines expectations and it comes with IDE support telling you which parameters you should specify.

How about introduce sub interface `ParameterizedSpecification`?
```java
	interface ParameterizedSpecification<T> extends Specification<T> {

		Map<String, Object> getParameters();

	}
```
Then you can do like this:
```java
	public static ParameterizedSpecification<User> userHasAgeLessThanParam(Integer age) {

		return new ParameterizedSpecification<>() {

			@Override
			public Predicate toPredicate(Root<User> root, CriteriaQuery<?> query, CriteriaBuilder cb) {
				return cb.lessThan(root.get("age").as(Integer.class), cb.parameter(Integer.class, "age"));
			}

			@Override
			public Map<String, Object> getParameters() {
				return Map.of("age", age);
			}

		};

	}

```

mp911de:
The beauty of `Specification` is that it can be used as lambda. Introducing another variant makes usage much more complex.

viftodi:
Hello,


> > We are using postges arrays in a function call with more than 100K arguments and this will help us to have smaller queries. It would be nice if it would be considered.
> 
> How does capturing parameters in a specification vs. passing these outside of the spec change your query structure? I think there's quite a mismatch in understanding.

Perhaps this is a particular use-case, but image a dynamic specification that is created say from a json.
And such specification is then applied with different parameters

> Considering the general mode of Spring SQL API's, there has always been a way to include parameters in-line or declare bind markers along with passing parameters as var-args or parameter source.
> 
> Allowing specifications to declare parameter placeholders falls exactly into this scheme. However, since specifications are all code and SQL is typically text, the proximity between the place that defines parameter expectations and the place that accepts parameters is quite far.

This might be true, but CriteriaBuilder allows the creation of parameter expression and EntityManager alllows you to bind them.
To me it can still be "close" in the code:

var spec = constructSpec();
var params = getParams();
repository.findAll(spec, params);

Or perhaps:
var spec = constructSpec();
for (var params: getMultipleParams()) {
 var res = repository.findall(spec, params);
 doSomething(res);
}

> Also, you cannot tell from a code snippet `repository.findOne(Queries.customersBySales(), …);` how many arguments the specification expects while a design such as `repository.findOne(Queries.customersBySales(tenant, salesThreshold));` clearly defines expectations and it comes with IDE support telling you which parameters you should specify.

This is true, and perhaps for simple specification it is not necessary to use parameters.

Usually when you can construct a Specification you can use CritieriaBuilder::equal(Expression<?> x, Object y)
And in hibernate by default this is optimized as a parameter so the query doesn't have to contain inline values.

But there are cases when you need to construct a complex specification, with perhaps tens of predicates that takes multiple arguments.
In our case the specification is very dynamic, the code basically "deserializes" it from a json model.

But to me the main interest in having parameters in specification is two-fold:

 1) To be able to express such Specifications but still use the high level repository API, in our particular case we are using:
CriteriaBuilder::function(String name, Class<T> type, Expression<?>... args)
And the only way you can pass arguments to a function call is via literals, and this case the query obviously contains the values, this might not be problematic in most cases but it can be  if the args are numerous or of a huge size (e.g. arrays in Postgres).
Another issue might be related with security as you might want in this particular case to guarantee parameters as to not leak values in the query. This can be expressed directly via CriteriaBuilder but you can't pass parameters if using a specification, you are forced to use directly the EntityManager and lose the higher level Specification and Repository API.

2) It might be desirable to separate the creation of a Specification from the execution and data binding.

Usually a specification is created say in the form getSpecForUsernameEqAndIdIn(String username, Long id) for example

But, and again this might be an edge case, but we have complex specifications that are actually created dynamically from another model. If you have the almost exact specification (only the values changes) then having parameters allows you to create the specification once, and execute it multiple times with different parameters.

But for our specific use case it's mainly the function argument that are problematic, we pass a huge array for example as a function call, and since it's a literal it is inlined in the query, so the construction of the query itself takes some time among other things. we have a func_call('{val1, val2, val3, ... val100_k'}) as a postgres array

This proposed API extension would allow for specific use cases to be able to execute a Specification that has  declared parameters, but still remains entirely backwards compatible.

Thank you

quaff:
> The beauty of `Specification` is that it can be used as lambda. Introducing another variant makes usage much more complex.

It will not break existing behavior, but add the ability to set parameter. `JpaSpecificationExecutor` will not be changed.

quaff:
> > Also, you cannot tell from a code snippet `repository.findOne(Queries.customersBySales(), …);` how many arguments the specification expects while a design such as `repository.findOne(Queries.customersBySales(tenant, salesThreshold));` clearly defines expectations and it comes with IDE support telling you which parameters you should specify.
> 
> How about introduce sub interface `ParameterizedSpecification`?
> 
> ```java
> 	interface ParameterizedSpecification<T> extends Specification<T> {
> 
> 		Map<String, Object> getParameters();
> 
> 	}
> ```
> 
> Then you can do like this:
> 
> ```java
> 	public static ParameterizedSpecification<User> userHasAgeLessThanParam(Integer age) {
> 
> 		return new ParameterizedSpecification<>() {
> 
> 			@Override
> 			public Predicate toPredicate(Root<User> root, CriteriaQuery<?> query, CriteriaBuilder cb) {
> 				return cb.lessThan(root.get("age").as(Integer.class), cb.parameter(Integer.class, "age"));
> 			}
> 
> 			@Override
> 			public Map<String, Object> getParameters() {
> 				return Map.of("age", age);
> 			}
> 
> 		};
> 
> 	}
> ```

I've created https://github.com/spring-projects/spring-data-jpa/pull/3209.

viftodi:
Hello,

I have given more thoughts to your questions and I wanted to clarify some things.

I can't talk for every user of the Specification API, but to me the CriteriaBuilder API is the JPA equivalent of JOOQ/QuerySQL or similar APIs. It allowed a type-safe API that allows you to represent almost any JPA query.

Obviously if a user can represent their query using JPQL and/or a spring data method they would do so.
Specifications / CriteriaBuilder is a lower level API, and is for certain use cases that usually require dynamic queries.

And this comes back to your question related to IDE support and figuring out from the API how many arguments you require.
This is the case for a JPQL query that has static parameters and even a spring data method.
But it is still a runtime error, a user can declare a JPQL query or a spring data method and forget to add a parameter and figure it out later at runtime.

Related to the call point of such an API, to me it's related to the dynamic nature of a Specification, if you have dynamic specifications, then your parameters and/or other values are dynamic as well.

But for simple specifications, the ParametersValues can be extended with multiple of methods.
Technically ParametersValue can be replaced by a Map<String, Value>, I thought initially that I'd require more of an abstraction compared to Map but ultimately it didn't require it.
One advantage other than API/documentation related to not using Map might be also by making ParametersValues immutable.
But if you think Map<String, Object> might be clearer, especially thanks to Map.of and a richer API, then I can handle a refactor.

But  lets assume repository.findAll(spec, ParametersValues.of("param1", val1, "param2", val2")
It won't solve the auto-completion and IDE help, but it will approach the clarity of which parameters are passed near the call site.

But again, to me Specification are used for more dynamic queries already, and this extension would allow even a more dynamic aspect of CriteriaBuilder.
It remains backwards compatible, and perhaps for most users they won't need parameters as well.

But to me the use cases are legitimate for two reasons:

1) Parameters can already be declared with current Specification (Because of CriteriaBuilder) but you can't set the parameters via JPASpecificationExecutor (Repository) so you can't execute them.,
2) This is debatable, but it might be beneficial or desirable to create some decoupling between the "query"/Specification itself, and the data you execute. (See my example of construct a specification in a way where you don't yet have access to the parameters, or when you which to construct it once, and execute it multiple times with multiple values, vs reconstructing the Specification as the values are part of the specification)
3. This is perhaps an edge case, but it is not always possible to have automatic parameters in queries.
In hibernate by default (hibernate.criteria.value_handling_mode is BIND) it will create parameters and bind them if you pass values via CriteriaBuilder e.q equal(expression, Object) but this isn't the case when you are passing a literal.
And in the particular case when using a function call you can only pass literals.

To me all this would do is align the API to be able to execute Specifications that you can't otherwise.

You can currently create specifications that contain parameters, and you will get an exception when running them since you can't set the parameters via the current API.

I hope this convinces you of the legitimacy of this request and I remain available for any further questions.

Thank you.

mp911de:
We discussed the topic as team and concluded that separation of the query and its parameters isn't a great programming model. 

For safe SQL query execution, the query declaration and the parameters are typically split (`SELECT foo FROM bar where baz = :baz`) as that is the mode which queries are build and parameters are passed. In arrangements where queries are colocated with parameters for execution (e.g. `jdbcTemplate.execute("SELECT … where baz = :baz", myBaz)`), you can immediately realize the relationship and parameter requirement.

Applying the same idea to specifications:

```java
var spec = constructSpec();

repository.findAll(spec, params);
```

comes around with an opaque specification that makes it difficult to determine which parameters are required.

We've been advocating a captive programming model for specifications in the following form:

```java
var spec = userHasFirstname("Foo").or(userHasLastname("Bar")).or(userHasLastname("Baz"));
repository.findAll(spec);
```

This approach keeps the specification together required parameters (improved locality). It avoids parameter clashes when using the same specification multiple times, so the need to come up with a parameter name that doesn't clash isn't necessary in the first place. For indexed parameters, offsetting has always been error-prone.

For reusable and improved repository usage, we recommend moving your method calls to inside the repository, by either using default interface methods or custom repository fragments.

That being said, thanks a lot for your suggestion. We do not want to move forward with this pull request and close it without merge.

quaff:
> We discussed the topic as team and concluded that separation of the query and its parameters isn't a great programming model.
> 
> For safe SQL query execution, the query declaration and the parameters are typically split (`SELECT foo FROM bar where baz = :baz`) as that is the mode which queries are build and parameters are passed. In arrangements where queries are colocated with parameters for execution (e.g. `jdbcTemplate.execute("SELECT … where baz = :baz", myBaz)`), you can immediately realize the relationship and parameter requirement.
> 
> Applying the same idea to specifications:
> 
> ```java
> var spec = constructSpec();
> 
> repository.findAll(spec, params);
> ```
> 
> comes around with an opaque specification that makes it difficult to determine which parameters are required.
> 
> We've been advocating a captive programming model for specifications in the following form:
> 
> ```java
> var spec = userHasFirstname("Foo").or(userHasLastname("Bar")).or(userHasLastname("Baz"));
> repository.findAll(spec);
> ```
> 
> This approach keeps the specification together required parameters (improved locality). It avoids parameter clashes when using the same specification multiple times, so the need to come up with a parameter name that doesn't clash isn't necessary in the first place. For indexed parameters, offsetting has always been error-prone.
> 
> For reusable and improved repository usage, we recommend moving your method calls to inside the repository, by either using default interface methods or custom repository fragments.
> 
> That being said, thanks a lot for your suggestion. We do not want to move forward with this pull request and close it without merge.

I've created https://github.com/spring-projects/spring-data-jpa/pull/3448 to handle this, WDYT? @viftodi

PR #3209, opened 2023-10-26
Introduce ParameterizedSpecification

quaff:
Add the ability to set parameter values for Specification

See GH-3192


mp911de:
Closing as per https://github.com/spring-projects/spring-data-jpa/pull/3192#issuecomment-1797953668

PR #3212, opened 2023-10-26
Use version-agnostic Hibernate presence check in `QueryEnhancerFactory`

AurMario:
Hibernate 5.6 supports JPA 3.0, and should be usable with Spring Data JPA 3.1.x. Currently, only Hibernate 6 is detected and allows for the HQL Parser usage. This PR consists in enabling the HQL Parser in the case of Hibernate 5 too.

- [x ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [x ] You submit test cases (unit or integration tests) that back your changes.
- [x ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
@@ -37,13 +37,16 @@ public final class QueryEnhancerFactory {
 	private static final boolean hibernatePresent = ClassUtils.isPresent("org.hibernate.query.TypedParameterValue",
 			QueryEnhancerFactory.class.getClassLoader());
 
+	private static final boolean hibernate5Present = ClassUtils.isPresent("org.hibernate.jpa.TypedParameterValue",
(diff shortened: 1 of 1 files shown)


mp911de:
After upgrading to a new major version, we typically require the new baseline. We can support Hibernate 5.6 on a best-effort basis without actually guaranteeing that everything is working, as our tests use the JPA 3.1 baseline.

We're fine with this one change, however, we do not intent to put in more effort into downgrading compatibility.

mp911de:
We've rewritten the presence detection for Spring Data JPA 3.2 already to a version-agnostic form and I polished up the pull request in the same direction. Can you also test whether the current 3.2 snapshots work with your Hibernate 5.6 arrangement?

mp911de:
Thank you for your contribution. That's merged and polished now into the `3.1.x` branch.

PR #3215, opened 2023-10-29
Change the BeanDefinitionNames constant interface to a final class Constant class that cannot be inherited.

ch4570:
The BeanDefinitionNames interface is a constant interface, so there is a risk of namespace pollution in other classes if they accidentally implement it. To prevent the risk of accidental implementation, we have transformed the constant interface into a final class with three constants. Additionally, we have updated the AuditingBeanDefinitionParser class, which currently lacks static imports, to include them.


mp911de:
`BeanDefinitionNames` is a package-private interface by design so there's not much risk to implement that interface for outside users.

PR #3232, opened 2023-11-19
Fixed the URL for the Spring Data Commons documentation

runbing:
I fixed an incorrect URL in the **antora.yml** file that was preventing other pages from linking to the Spring Data Commons documentation.


mp911de:

Issue #3263, opened 2023-12-08
JSqlParserQueryEnhancer removes WITH clauses for paginated and sorted native queries


lumi77:
Repository class:
```java
@Repository
public interface ProjectsRepository extends JpaRepository<Project, Integer> {
    String QUERY = """
            WITH all_projects AS
            (
                SELECT *
                FROM projects
            )
            SELECT * FROM all_projects
            """;

    @Query(value = QUERY, nativeQuery = true)
    List<Project> findAllNativeQuery();

    @Query(value = QUERY, nativeQuery = true)
    Page<Project> findAllNativeQuery(Pageable pageable);
}
```

Test class for the repository
```java
@DataJpaTest
class ProjectsRepositoryTest {
    @Autowired
    private ProjectsRepository projectsRepository;

    @Test
    void findAll() {
        var actual = projectsRepository.findAllNativeQuery();

        assertThat(actual).isEmpty();
    }

    @Test
    void findAllPageableUnsorted() {
        var actual = projectsRepository.findAllNativeQuery(PageRequest.of(0, 10));

        assertThat(actual.getContent()).isEmpty();
    }

    @Test
    void findAllPageableSorted() {
        var actual = projectsRepository.findAllNativeQuery(PageRequest.of(0, 10, Sort.Direction.ASC, "name"));

        assertThat(actual.getContent()).isEmpty();
    }
}
```
When `JSqlParser` is not on the class path, all three tests pass.

When `JSqlParser` is on the class path, the third tests fails with an exception:
```
org.springframework.dao.InvalidDataAccessResourceUsageException: could not prepare statement [Table "ALL_PROJECTS" not found; SQL statement:
SELECT * FROM all_projects ORDER BY name ASC fetch first ? rows only [42102-224]] 
[SELECT * FROM all_projects ORDER BY name ASC fetch first ? rows only]; SQL [SELECT * FROM all_projects ORDER BY name ASC fetch first ? rows only].
```
Both the printed query `"SELECT * FROM all_projects ORDER BY name ASC fetch first ? rows only"` and the error message `"Table "ALL_PROJECTS" not found"` imply that the WITH clause 
 ```sql
WITH all_projects AS
            (
                SELECT *
                FROM projects
            )
```
has been lost while preprocessing the query in `JSqlParserQueryEnhancer`. This happens in `public String applySorting(Sort sort, @Nullable String alias)`

[...]

PR #3264, opened 2023-12-11
Render entire statement after applying sort. 

christophstrobl:
This PR makes sure to render the entire statement after applying the sort expression via the `QueryEnhancer`.
Previously only parts, the actual statement body, had been considered.

Fixes: #3263

### DIFF
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryEnhancerUnitTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryEnhancerUnitTests.java
@@ -179,6 +179,17 @@ class QueryEnhancerUnitTests {
 		assertThat(query).endsWithIgnoringCase("ORDER BY p.firstname, p.lastname asc");
 	}
 
+	@Test // GH-3263
+	void doesNotDropWithExpressionWhenApplySort() {
+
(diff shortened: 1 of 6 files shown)


spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryEnhancerUnitTests.java
```diff
@@ -179,6 +179,17 @@ void findsExistingOrderByIndependentOfCase() {
 		assertThat(query).endsWithIgnoringCase("ORDER BY p.firstname, p.lastname asc");
 	}
 
+	@Test // GH-3263
+	void doesNotDropWithExpressionWhenApplySort() {
```
quaff, line 183: It's better to use `Sorting` instead of `Sort` for consistency.

mp911de:
Looks good, feel free to merge this one.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/QueryEnhancerUnitTests.java
christophstrobl: thanks @quaff - additionally I'm not happy with the `doesNotDrop` part. will rename the method.

Issue #3274, opened 2023-12-18
Generic @MappedSuperClass class doesn't works since 3.0


quaff:
It works fine with 2.7 but failed with 3.0.
```
Caused by: java.lang.IllegalArgumentException: Unable to locate Attribute with the given name [ringIndex] on this ManagedType [java.lang.Object]
	at org.hibernate.metamodel.model.domain.AbstractManagedType.checkNotNull(AbstractManagedType.java:225) ~[hibernate-core-6.3.1.Final.jar:6.3.1.Final]
	at org.hibernate.metamodel.model.domain.AbstractManagedType.getAttribute(AbstractManagedType.java:148) ~[hibernate-core-6.3.1.Final.jar:6.3.1.Final]
	at org.hibernate.metamodel.model.domain.AbstractManagedType.getAttribute(AbstractManagedType.java:43) ~[hibernate-core-6.3.1.Final.jar:6.3.1.Final]
	at org.springframework.data.jpa.repository.query.QueryUtils.requiresOuterJoin(QueryUtils.java:836) ~[spring-data-jpa-3.2.0.jar:3.2.0]
```
The hibernate team suggest it's should be fixed at Spring Data JPA side.
>> the JPA specification requires that generic attributes are reported as generic-ly typed. The resolution of an attribute’s concrete type is an internal mechanism used to make Hibernate queries work. Your change is not correct, and the issue described in this Jira is on Spring’s side so I suggest reporting it there.

>> As with Java's reflection API, EntityType.getAttribute() gives you a representation of the inherited attribute as it was declared in the supertype. It does not do substitution of generic type arguments from the inheriting subclass, and I don't see any support anywhere in the JPA API or specification which hints that it should.So I believe that it's correct for Hibernate to return Object.class here, and I think this PR should be closed.

see https://hibernate.atlassian.net/browse/HHH-16144 and https://github.com/hibernate/hibernate-orm/pull/7630

christophstrobl:
Thanks @quaff for reporting and taking the time to create a PR. We'll have a look.

PR #3275, opened 2023-12-18
Do not rely on ManagedType by default

quaff:
ManagedType may be erased type if the attribute is declared as generic, take Hibernate 6.x for example, exception is thrown like:
```
java.lang.IllegalArgumentException: Unable to locate Attribute with the given name [name] on this ManagedType [java.lang.Object]
	at org.hibernate.metamodel.model.domain.AbstractManagedType.checkNotNull(AbstractManagedType.java:225) ~[hibernate-core-6.3.1.Final.jar:6.3.1.Final]
	at org.hibernate.metamodel.model.domain.AbstractManagedType.getAttribute(AbstractManagedType.java:148) ~[hibernate-core-6.3.1.Final.jar:6.3.1.Final]
	at org.hibernate.metamodel.model.domain.AbstractManagedType.getAttribute(AbstractManagedType.java:43) ~[hibernate-core-6.3.1.Final.jar:6.3.1.Final]
	at org.springframework.data.jpa.repository.query.QueryUtils.requiresOuterJoin(QueryUtils.java:836) ~[spring-data-jpa-3.2.0.jar:3.2.0]
```

Fix GH-3274

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -81,6 +81,7 @@ import org.springframework.util.StringUtils;
  * @author Vladislav Yukharin
  * @author Chris Fraser
  * @author Donghun Shin
+ * @author Yanming Zhou
  */
 public abstract class QueryUtils {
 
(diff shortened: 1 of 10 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -821,20 +822,28 @@ private static boolean requiresOuterJoin(From<?, ?> from, PropertyPath property,
...
+			if (managedType != null) {
+				propertyPathModel = (Bindable<?>) managedType.getAttribute(segment);
+			} else {
+				propertyPathModel = from.get(segment).getModel();
+			}
 		} else {
+			// ManagedType may be erased type for some vendor if the attribute is declared as generic
+			// see: https://hibernate.atlassian.net/browse/HHH-16144
+			// see: https://github.com/hibernate/hibernate-orm/pull/7630
+			// see: https://github.com/jakartaee/persistence/issues/562
```
christophstrobl, line 846: I'm curious if we could hide the vendor specifics behind eg. the `PersistenceProvider` so we can encapsulate different behaviours when getting hold of the attribute .

I'm thinking of something that would allow us to do something like the following:
```java
Bindable<?> propertyPathModel = PersistenceProvider.segmentResolver(from).resolve(segment);
``` 
Having `PersistenceProvider` implement a let's call it `PathSegmentResolver` defaulting to `from.get(segment).getModel()`, we could move the eclipselink specific requirements with all the type checks to the ECLIPSELINK implementation.

quaff: How to resolve `PersistenceProvider` in method `QueryUtils::requiresOuterJoin`?

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
```diff
@@ -821,20 +822,28 @@ private static boolean requiresOuterJoin(From<?, ?> from, PropertyPath property,
...
 			propertyPathModel = from.get(segment).getModel();
```
mp911de, line 847: `from.get("parent").getModel()` returns the erased type instead of the correct one inferred from type parameters.

mp911de:
This isn't a working fix, it just adds some additional complexity that happens to work by accident and not by intent.

PR #3280, opened 2023-12-20
Fix JPQL parser issues with numeric values

christophstrobl:
Fixes: #3277 
Fixes: #3308


spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JpqlComplianceTests.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2023 the original author or authors.
```
mp911de, line 2: Nit: 2024

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/SqlParserTests.java
```diff
@@ -0,0 +1,113 @@
...
+
+import java.util.List;
+
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author Christoph Strobl
+ */
+public abstract class SqlParserTests {
```
mp911de, line 28: How about `JpqlParserAndRenderTests`?

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/JpqlComplianceTests.java
christophstrobl: 3 weeks ago this had been valid ;) - I'll change after rebase.

PR #3281, opened 2023-12-20
Fixes FileSystemException

cagliostro92:
image  

Under Windows operating system, unfortunately, I'm unable to build the JPA project because of a FileSystemException thrown at the act of delete the test-eclipselink-meta.log file in the EclipseLinkMetaAnnotatedQueryMethodIntegrationTests class. The proposal is to use a temporary file whose life cycle can be easily managed by the underlying operating system. Thanks in advance to the JPA team.

### DIFF
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
@@ -84,7 +84,8 @@ class EclipseLinkMetaAnnotatedQueryMethodIntegrationTests {
 		LOG_FILE = logFile.toAbsolutePath().toString();
 	}
 
-	@AfterAll static void deleteLogFile(){
+	@AfterAll
+	static void deleteLogFile(){
 		logFile.toFile().deleteOnExit();
 	}
 
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
@@ -31,7 +31,6 @@ import javax.sql.DataSource;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
```diff
@@ -74,16 +75,12 @@ class EclipseLinkMetaAnnotatedQueryMethodIntegrationTests {
...
-		new FileOutputStream(LOG_FILE).close();
-	}
-
-	@AfterAll
-	static void deleteLogfile() throws IOException {
-		FileSystemUtils.deleteRecursively(Path.of(LOG_FILE));
+	@BeforeAll
+	static void createLogfile() throws IOException {
+		Path logFile = Files.createTempFile("test-eclipselink-meta", ".log");
+		LOG_FILE = logFile.toAbsolutePath().toString();
```
christophstrobl, line 83: how about making use of junits `@TempDir` leveraging automatic cleanup.

christophstrobl:
thank you @EdoardoP92 for bringing this up and taking the time to create a PR.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
EdoardoP92: Hi, thank you for the reply. Unfortunately with `@TempDir` there is the same problem...  

```java
@TempDir
static Path TMP;

    @BeforeAll
    static void createLogfile() throws IOException {
        Path logFile = TMP.resolve("test-eclipselink-meta.log");
        LOG_FILE = logFile.toAbsolutePath().toString();
    }
```

image

EdoardoP92:
> thank you @EdoardoP92 for bringing this up and taking the time to create a PR.

@christophstrobl  thank you too

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/EclipseLinkMetaAnnotatedQueryMethodIntegrationTests.java
mp911de: Generally switching to `@TempDir` is generally a good thing but we cannot fix the underlying problem that Eclipselink doesn't close the underlying logger. The offending code is in `EntityManagerSetupImpl`.

EdoardoP92: @mp911de Let me know how I can resolve this issue, if my proposal to use temporary files is ok or if you prefer some other changes. I want to contribute, but I can't check my contribution without run the entire test suite. In the meantime, after further analysis, I can for sure open an issue, or a PR to the EclipseLink project. Thank you very much.

PR #3284, opened 2023-12-28
Refine AspectJ plugin build phase config

kriegaex:
This is a follow-up on #3282, where a reviewer's misunderstanding (running test builds with wrong Maven phases) led to parts of the PR getting falsely erased. This PR ~~not only reinstates those erased parts, but also~~ enables developers who do not know the build configuration well enough to know that they should run `mvn process-classes` rather than `mvn compile` to now use the latter. This new convenience feature is not a fix or change of the previous PR, but changes a build phase which has been there long before the PR.

@mp911de, @odrotbohm

### DIFF
--- a/spring-data-jpa/pom.xml
+++ b/spring-data-jpa/pom.xml
@@ -360,8 +360,7 @@
 			<!--
 				Attention: aspectj-maven-plugin MUST be declared AFTER maven-compiler-plugin, if they are both supposed to run
 				in the same default phases 'compile' and 'test-compile', but execution order is to be guaranteed. Then, you have
-				the convenience of being able to run e.g. 'mvn [clean] compile' instead of 'mvn [clean] process-classes'. For a
-				slightly less convenient, but less refactoring-error-prone solution, see the commented-out phases below.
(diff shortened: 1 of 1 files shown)


kriegaex:
@mp911de, like I said over there in the other PR, after your ANTLR optimisation and removal of Replacer plugin, I dropped the first commit, leaving only the AspectJ Maven phase change to default. I still think, this would help more than it would hurt. It is documented well enough to avoid anyone messing up in the future. And if they wouldmess up, they would notice immediately, because the build would break.

mp911de:
Alright, we're on the same page now. Care to remove the commented-out XML bits so I can merge the PR without polishing it? Happy to keep the explanatory comment though.

kriegaex:
Actually, I left them there on purpose, and the comment even points to them. I like things to be explicit rather than expecting other developers to possess sacred knowledge. But if you insist, of course I can remove them, as soon as I am near my workstation again. 🙂

kriegaex:
OK, done, even though I still think that keeping the phases and the hint as a help for the future would be better. You can either merge the whole PR with both commits or cherry-pick just the first one, if you change your mind.

kriegaex:
Relates to https://github.com/antlr/antlr4/issues/4504.

PR #3301, opened 2024-01-04
Allow Supertypes Specifications in JpaSpecificationExecutor

### LINKED ISSUE #3300: Allow Supertypes Specifications in JpaSpecificationExecutor
See the following commit: https://github.com/spring-projects/spring-data-jpa/commit/4c4512578ebe6333cc7c388e8533e11c7e0a396f
I added a test in your codebase which is exactly the same as I am going to describe. 

Suppose we have an abstract `@MappedSupperclass` `AbstractMappedType` and a concrete entity `ConcreteType1` extending `AbstractMappedType`. 

Then we also have `MappedTypeRepository` which is a `@NoRepositoryBean` but does extend `JpaSpecificationExecutor` and must therefore look like this: 

```java
@NoRepositoryBean
public interface MappedTypeRepository<T extends AbstractMappedType> extends JpaRepository<T, Long>, JpaSpecificationExecutor<T> {
}
``` 

However when using this repository in its abstract form, we face some issues. 

```java 
@Autowired
@Qualifier("concreteRepository1")
private MappedTypeRepository<?> mappedTypeRepository;

@Test
void testUseMappedTypeRepository() {
  concreteRepository1.save(new ConcreteType1("Test"));

  List<? extends AbstractMappedType> findings = mappedTypeRepository.findAll(
      where(mappedTypeAttribute1Equals("Test")));
  assertThat(findings).isNotEmpty();
}

private static Specification<AbstractMappedType> mappedTypeAttribute1Equals(String attribute1) {
  return (root, query, criteriaBuilder) -> criteriaBuilder.equal(root.get(AbstractMappedType_.ATTRIBUTE1), attribute1);
}
```

This does not work because `Specification<AbstractMappedType>` cannot be applied as first parameter for 
`List<T> findAll(Specification<T> spec);` of `JpaSpecificationExecutor<T>`.

IMO, the method `JpaSpecificationExecutor#findAll` should be defined as follows, accepting also specifications of supertypes of `T`. 
```java
List<T> findAll(Specification<? super T> spec); 
``` 

In fact, I have already implemented change, and you can have a look at #3301

aykborstelmann:
Fixes #3300 

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/convert/QueryByExamplePredicateBuilder.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/convert/QueryByExamplePredicateBuilder.java
@@ -93,7 +93,7 @@ public class QueryByExamplePredicateBuilder {
 	 * @return {@literal null} indicates no constraints
 	 */
 	@Nullable
-	public static <T> Predicate getPredicate(Root<T> root, CriteriaBuilder cb, Example<T> example,
+	public static <T> Predicate getPredicate(Root<? extends T> root, CriteriaBuilder cb, Example<T> example,
 			EscapeCharacter escapeCharacter) {
 
 		Assert.notNull(root, "Root must not be null");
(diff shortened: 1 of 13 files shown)


mp911de:
We discussed the issue within our team and we don't want to update our code to consider type hierarchy changes across specifications.

PR #3321, opened 2024-01-18
Improve string query parsing flow. 

christophstrobl:
Resolves: #3310, #3311

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
@@ -18,6 +18,8 @@ package org.springframework.data.jpa.repository.query;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.Query;
 
+import java.util.Objects;
+
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.jpa.repository.QueryRewriter;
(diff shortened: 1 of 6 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
```diff
@@ -179,4 +188,77 @@ protected String potentiallyRewriteQuery(String originalQuery, Sort sort, @Nulla
...
+	}
+
+	/**
+	 * Value object with optimized {@link Object#equals(Object)} to cache a query based on its query string and
+	 * {@link Sort sorting}.
+	 *
+	 * @since 3.2.3
+	 * @author Christoph Strobl
+	 */
+	static class CachableQuery {
```
mp911de, line 213: The caching key is already quite complex. I'd suggest to keep only the `isUnsorted` code path and refine the way we render queries with a sort. That is, render everything before and after `ORDER BY` in a first pass (e.g. held in a `QueryEnhancer` instance). Then, render `Sort` with rather simple string concatenation.

christophstrobl: The above is deliberately targeting the 3.2.x line without applying major changes.

christophstrobl: Created #3326 to follow up on splitting the transformation (targeting _main_ development line)

PR #3322, opened 2024-01-19
Rename iteration variable name in PersistenceProviderUnitTests for readability

devholic22:
Rename iteration variable name in `PersistenceProviderUnitTests` for readability

In `PersistenceProviderUnitTests` `toResourcePaths` method, use `iface` in `interfacesToImplement`.

but in my opinion, `targetInterface` is better than `iface` for readability. (`iface` not means direct this variable is a unit of `interfacesToImplement`.)

In my guess, `interface` word is already used in java default words. (type name) so written in `iface`.

could you consider this suggestion?

(The two check boxes at the bottom do not seem to correspond to simple word changes, so I did not check them.)


mp911de:

PR #3330, opened 2024-01-23
QueryUtils do not remove all "order by" parts when making a count que…

### LINKED ISSUE #3329: QueryUtils derive broken count query when original one contains ORDER BY with new lines
Stumbled upon this issue while moving SQL queries to text blocks. Consider the following:
```
"""
SELECT * FROM table t
WHERE t.id = :id
  AND t.deleted = FALSE
ORDER BY t.active DESC,
         t.creation_date DESC
"""
```
Such a query will fail to run because Spring Data tries to derive a count query from this one and removes the `ORDER BY` part, but only when it's in a single line
```
QueryUtils.ORDER_BY_PART = "(?iu)\\s+order\\s+by\\s+.*";
```

If it's multi-line text, the count query turns like 
```
select count(1) FROM table t
WHERE t.id = :id
AND t.deleted = FALSE
t.creation_date DESC
```

and expectedly fails.

edudar:
…ry when they are spread across multiple lines

Closes #3329


christophstrobl:
thank you @edudar - merged to _main_ development line and back ported to maintenance branches.

PR #3346, opened 2024-02-02
Attempt to reproduce problem reported by spring-data team https://github.com/OpenFeign/querydsl/issues/275#issuecomment-1913725963

velo:
Hello team,

This PR is *not* meant to be merged by any means, I just want to see if spring CI will break and I can learn anything from it.

Have been having an interesting discussion with @odrotbohm on how the spring-data team thinks my new querydsl-jpa-spring module can break break the whole ecosystem: https://github.com/OpenFeign/querydsl/issues/275

But, neither myself nor @danielvion seem to be able to comprehend what is the actual issue.

By any chance can I get access to your internal discussions so I might understand what I'm missing?

So far the "feedback" I got is:

* "introducing Spring Data support in Querydsl in turn creates a cyclic dependency", I don't understand how!
* "is quite anti-idiomatic to the Spring Data repository concept", well, I just did the exact same thing as `@Query`, but using querydsl language.... not sure what is wrong here.

My new module allows to declare queries like this:

```
  default Customer findById(long id) {
    return select(C).from(C).where(C.id.eq(id)).fetchOne();
  }
```

This is not meant to be an alternative to `Customer findById(long id);`, but rather an alternative to the situations where you need a `@Query("select c from c where c.id = :id") `.  I know is not the case for the example, it's a very silly example where neither my solution nor `@Query` are the best option.

So far, the best lead we got, would be some sort of cyclic dependency going to happen between spring-bom and querydsl-bom.

Querydsl-bom is not opinionated about third party dependencies, all it provides is a unified version for querydsl-* , so the only way I can see a cyclic reference happening is if someone adds querydsl-jpa-spring  to spring-data.  But, since the team already considers this anti-idiomatic, I don't see it happening.

Can anyone provide us with a real example on how things actually break?  This very abstract conversation makes my head hurt and do not help identifying the issue.

I already proposed to @danielvion to delete the module if that becomes a blocker to moving things along, but I very much prefer to delete it with a understandable reason.

Thanks, and feel free to close this if not allowed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,7 @@
 		<postgresql>42.6.0</postgresql>
 		<springdata.commons>3.3.0-SNAPSHOT</springdata.commons>
 		<vavr>0.10.3</vavr>
+		<querydsl>6.0</querydsl>
 
 		<hibernate.groupId>org.hibernate</hibernate.groupId>
 
--- a/spring-data-envers/pom.xml
+++ b/spring-data-envers/pom.xml
@@ -68,10 +68,9 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.querydsl</groupId>
+			<groupId>io.github.openfeign.querydsl</groupId>
 			<artifactId>querydsl-jpa</artifactId>
 			<version>${querydsl}</version>
-			<classifier>jakarta</classifier>
 			<optional>true</optional>
 		</dependency>
 
--- a/spring-data-jpa/pom.xml
+++ b/spring-data-jpa/pom.xml
@@ -192,17 +192,15 @@
 
 		<!-- QueryDsl -->
 		<dependency>
-			<groupId>com.querydsl</groupId>
+			<groupId>io.github.openfeign.querydsl</groupId>
 			<artifactId>querydsl-apt</artifactId>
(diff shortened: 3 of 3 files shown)


mp911de:
The OpenFeign fork of Querydsl is going to go away, so we're not going to follow up on the issues caused by a foreign fork.

velo:
@mp911de ow, that is very interesting to know, since I'm the one maintaining the fork.

Is there anything else I need to know?

mp911de:
Communication across multiple project hops is always difficult, especially if not all folks are looped in that are affected. Querydsl folks have released just recently a release and signalled to us that they wanted to work with you to merge back some of your changes so that there's no need for another fork. Don't shoot the messenger.

Anyways, we do not intend to switch to another fork of Querydsl as that would lead to even more community fragmentation.

velo:
Ow, ok, I knew I was missing something.

Well, they indicated some wiliness to work with fork yes, but, they are concern about this supposed "cyclic dependency between Spring Boot, Spring Data & Querydsl" that Oliver reported.

So far, I didn't managed to find any evidence of the problem.

My hope was that your CI would be able to provide me more information.

Issue #3366, opened 2024-02-14
Fix typo in unused


GangEunzzang:
I found a typo in an unused comment within the codebase.

mp911de:
We use `//` to indicate our formatter a dedicated line break, so it is not a typo.

PR #3367, opened 2024-02-14
Remove unused typo.

GangEunzzang:
See: #3366

This PR deletes an unused comment in the codebase that contains a minor typo. The comment is not utilized, and its removal improves code cleanliness and readability.


mp911de:
We use `//` to indicate our formatter a dedicated line break, so it is not a typo.

PR #3368, opened 2024-02-14
Code Simplification: Removed unnecessary else clause in if statement

chs98412:
Hello! This PR aims to simplify the code by removing the unnecessary else clause in the if statement. Previously, the code had an if-else structure where it returned different objects based on the condition method.isModifyingQuery(). However, since the return statement is already present within the if block, the else clause can be safely removed without affecting the logic. The modified code is more concise and easier to read. Your review and feedback on the changes would be greatly appreciated. Thank you!


christophstrobl:
Thank you @chs98412 - merged to _main_ development line.

Issue #3349, opened 2024-02-02
Skip `JOIN` for predicates that compare the primary key of a `@ManyToOne` relationship


scordio:
Given the following entities:

```java
@Entity
class Author {

  @Id @GeneratedValue(strategy = GenerationType.AUTO)
  private Long id;

  private String name;

  // getters and setters omitted
}

@Entity
class Book {

  @Id @GeneratedValue(strategy = GenerationType.AUTO)
  private Long id;

  private String name;

  @ManyToOne(fetch = FetchType.LAZY)
  private Author author;

  // getters and setters omitted
}
```

and the following repository:

```java
interface BookRepository extends Repository<Book, Long> {

  Book save(Book book);

  List<Book> findAllByAuthorId(Long authorId);

}
```

`findAllByAuthorId` seems to trigger an unnecessary JOIN.

I initially noticed this behavior on Spring Boot 2.7.14 with DB2 for z/OS so I tried to reproduce it locally on newer Spring Boot versions.

Surprisingly, 3.0.x and 3.1.x seem to work properly, i.e., no unnecessary JOIN, while the JOIN is back again on 3.2.x and 3.3.x. That's why I mentioned it as a regression in the title.

Here are the derived queries for each Spring Boot version with an H2 database, captured with P6Spy:

### [2.7.18](https://github.com/scordio/spring-boot-playground/actions/runs/7766870886/job/21183160001#step:4:129)

```sql
select book0_.id as id1_1_, book0_.author_id as author_i3_1_, book0_.name as name2_1_ from book book0_ left outer join author author1_ on book0_.author_id=author1_.id where author1_.id=1;
```

### [3.0.13](https://github.com/scordio/spring-boot-playground/actions/runs/7766873160/job/21183164672#step:5:126)

```sql
select b1_0.id,b1_0.author_id,b1_0.name from book b1_0 where b1_0.author_id=1;
```

### [3.1.8](https://github.com/scordio/spring-boot-playground/actions/runs/7766873160/job/21183164745#step:5:124)
```sql
select b1_0.id,b1_0.author_id,b1_0.name from book b1_0 where b1_0.author_id=1;
```

### [3.2.2](https://github.com/scordio/spring-boot-playground/actions/runs/7766873160/job/21183164815#step:5:124)
```sql
select b1_0.id,b1_0.author_id,b1_0.name from book b1_0 left join author a1_0 on a1_0.id=b1_0.author_id where a1_0.id=1;
```

[...]

Issue #3307, opened 2024-01-09
Unable to locate Attribute with the given name on parametrized supertype


HVollmilch:
Hi there,

during the update to spring boot 3.x i encountered some issues related to generic types used in entities. I reported a bug in hibernate-core ([Jira](https://hibernate.atlassian.net/jira/software/c/projects/HHH/issues/HHH-17405)), which was fixed as part of spring boot 3.2.1 (hibternate 6.4.1) but now my jpa repositories are the next ones, that are not happy with my implementation. 

Lets assume we have 2 abstract classes Parent and Child, that know each other

```
@Data
@MappedSuperclass
public abstract class Parent<C extends Child<?>> {

    @OneToMany(mappedBy = "parent", cascade = CascadeType.ALL, fetch = FetchType.LAZY, orphanRemoval = true)
    private List<C> children = new LinkedList<>();

    public abstract Long getId();

    public abstract void setId(Long id);

}

@Data
@MappedSuperclass
public abstract class Child<P extends Parent<?>> {

    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

    @ManyToOne(optional = false)
    private P parent;

}
```

And we have multiple implementations of them

```
@Data
@Entity
public class ParentA extends Parent<ChildA> {

    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

}

@Data
@Entity
public class ParentB extends Parent<ChildB> {

    @Id
    private Long id;

}

@Data
@Entity
public class ChildA extends Child<ParentA> {
}

@Data
@Entity
public class ChildB extends Child<ParentB> {
}

```

The ids of the parents are not specified in the abstract class cause i need different generation strategies depending on the specific type.

And then we have a repository interface for children, all specific-type-repositories are inheriting:

```
@NoRepositoryBean
public interface ChildRepository<C extends Child<?>> extends JpaRepository<C, Long> {

    boolean existsByIdAndParentId(Long id, Long parentId);

}
```

I will work with this ChildRepository in an abstract service.

If i try to execute the "extistsByIdAndParentId" method, i'll get the following exception
`Caused by: java.lang.IllegalArgumentException: Unable to locate Attribute with the given name [id] on this ManagedType [com.hibernate.data.core.Parent]`

But this kind of inheritance worked in my previous version using spring boot version 2.6.1

I've attached a application of this simplified usecase, where you can recreate my problem
[...]

PR #3374, opened 2024-02-19
Use implicit join to create Predicate if possible

quaff:
Fix GH-3349


quaff:
It should fix GH-3274 and GH-3307 also.

christophstrobl:
Thank you @quaff! Would you mind adding some tests to verify the behaviour?

quaff:
> Thank you @quaff! Would you mind adding some tests to verify the behaviour?

I'm trying.

quaff:
@christophstrobl Could you point out which existing test assert on generated sql?

quaff:
> > Thank you @quaff! Would you mind adding some tests to verify the behaviour?
> 
> I'm trying.

@christophstrobl Done!

quaff:
see https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#hql-implicit-join

quaff:
There is a subtle change by this PR, since implicit join will use inner join, then `IS_NULL` will require `ManyToOne` relation entity exists but property of that entity is null, before this commit, it imply relation entity is null or property of entity is null. It's easy to keep the behavior as before, but I doubt it's right to do that.

Issue #3388, opened 2024-03-01
Add CI build against Hibernate 6.2


mp911de:
Hibernate 6.2.x is a LTS version that we want to verify our build against.

PR #3390, opened 2024-03-04
Add build profile for hibernate 6.2.x

christophstrobl:
Resolves: #3388 
should also be back ported to _3.2.x_ (_3.1.x_ is based on hibernate 6.2).


pom.xml
```diff
@@ -69,6 +70,18 @@
...
 			</repositories>
 		</profile>
+		<profile>
+			<id>hibernate-62</id>
+			<properties>
+				<hibernate>${hibernate-62}</hibernate>
+			</properties>
+			<dependencies>
+				<dependency>
+					<groupId>jakarta.persistence</groupId>
```
mp911de, line 80: What breaks if you omit the JPA spec?

christophstrobl: the querydsl plugin

mp911de:

PR #3400, opened 2024-03-15
Replace links to the Java EE JPA Spec with ones to the Jakarta Persistence Spec

AlexElin:
Replace links to the Java EE JPA Spec with ones to the Jakarta Persistence Spec.

Now the spec is released as Jakarta, not Java EE. That's why I've updated the links.


mp911de:
Thank you for the suggestion. With Oracle's documentation variant, there was only a PDF available. Now with Jakarta EE, we have also a HTML variant available. Care to revise your links into https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1? HTML gives us the advantage of working deep links.

AlexElin:
Hi @mp911de 
I've revised the links the way you had suggested. Pls, take a look

mp911de:

PR #3408, opened 2024-03-24
Improved consistency in writing Util classes and removed raw types to make types more stable

ch4570:
Hello.
I've been putting your `Spring Data JPA` to good use in my practice.

I've modified & improved the code of a few classes to create a PR and upload it.
Below are the classes I worked on and the reasons for the changes.

- `PageableUtils.java`
The abstract keyword is used to emphasize that the utility class does not need to be instantiated.

- `Meta.java`
We changed values in the Meta class because it is a private field that is only used internally and it is better to initialize it with a LinkedHashMap beforehand instead of checking if the element is empty every time setValue is called and initializing it with a LinkedHashMap.

- `NamedQuery.java`
Since EntityManager implements the AutoCloseable interface, it is better to use try-with-resources rather than explicitly calling close in the finally block after using resources, so we modified the code.

- `JsqlParserUtils.java`
Since the JSqlPaserUtils class is also a Util class, we clarified the semantics with the abstract keyword and modified the private constructor to throw the same exception to be consistent with PageableUtils.

The getJsqlCount and getJsqlLower methods are using ExpressionList as a raw type, but ExpressionList only allows classes that implement the Expression interface, and the withParameters method of Function class is only used by JsqlParserUtils, so we changed it to a type-safe code that uses Generic.

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -1,2 +1,2 @@
-#Wed Oct 04 16:58:05 PDT 2023
-distributionUrl=https\://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.5/apache-maven-3.9.5-bin.zip
+#Thu Dec 14 08:40:35 CET 2023
+distributionUrl=https\://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.6/apache-maven-3.9.6-bin.zip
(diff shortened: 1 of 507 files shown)


christophstrobl:
Thank you @ch4570.

ch4570:
@christophstrobl Thank you for merging my PR.
I'm wondering why I don't see any contributor indication when the code is merged into the main branch.

christophstrobl:
@ch4570 Thank you for the contribution - the email used does not seem to be linked to your account (like in this commit c872085d367f9f551a2478b4ae3bcb5c5e151ee1 from the original PR).

ch4570:
@christophstrobl I didn't realize I had a different username because I committed from work.
I apologize for bothering you with this question.

Thank you for your kind response.
I'll try to contribute more in the future. 🥇

PR #3414, opened 2024-04-02
Code Simplification: Removed unnecessary else clause in if statement in AbstractJpaQuery, EqlQueryRenderer, JpaEntityInformationSupport

chs98412:
Hello! This PR aims to simplify the code by removing the unnecessary else clause in the if statement. While I was looking through `AbstractJpaQuery`, `EqlQueryRender`, and `JpaEntityInformationSupport`, I discovered an unnecessary `else` syntax and suggested pr similar to before. I think the `else` clauses can be safely removed without affecting the logic. Your review and feedback on the changes would be greatly appreciated. Thank you!


mp911de:
Thanks for your suggestion. The affected code uses consistently `if … if else … else` styling and we want to keep it that way.

PR #3419, opened 2024-04-09
Improved time complexity to O(N) in the second method.

dukbong:
- [  ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

#### Comment
I have updated the time complexity of the method from O(N^2) to O(N).
Additionally, all the test cases in the mapping package have passed successfully.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/mapping/JpaMetamodelMappingContext.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/mapping/JpaMetamodelMappingContext.java
@@ -171,25 +171,24 @@ public class JpaMetamodelMappingContext
 		 */
 		@Nullable
 		private Metamodel getMetamodelFor(Class<?> type) {
-
-			for (Metamodel model : metamodels) {
-
-				try {
+			Metamodel currentModel = null;
+			try {
+				for (Metamodel model : metamodels) {
+					currentModel = model;
 					model.managedType(type);
 					return model;
-				} catch (IllegalArgumentException o_O) {
-
-					// Fall back to inspect *all* managed types manually as Metamodel.managedType(…) only
-					// returns for entities, embeddables and managed superclasses.
-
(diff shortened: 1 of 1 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/mapping/JpaMetamodelMappingContext.java
```diff
@@ -171,25 +171,24 @@ public boolean isMetamodelManagedType(Class<?> type) {
...
-						if (type.equals(managedType.getJavaType())) {
-							return model;
-						}
+				}
+			} catch (IllegalArgumentException o_O) {
+				
+				// Fall back to inspect *all* managed types manually as Metamodel.managedType(…) only
+				// returns for entities, embeddables and managed superclasses.
+				
+				for (ManagedType<?> managedType : currentModel.getManagedTypes()) {
```
christophstrobl, line 186: Looks like we may run into a potential NPE here.

dukbong: Thanks! @christophstrobl 

I made a mistake in adding the null check. 😢 
Based on your feedback, I have now corrected the null check.

dukbong: Hi! @christophstrobl 

Upon reevaluation, I believe there shouldn’t be an NPE occurrence as you mentioned.

To meet the conditions for an NPE as mentioned in the feedback, ‘metamodels’ within the try block would need to be null. However, since the constructor already handles this check, I anticipate no NPE occurrence within the catch block.

Note: There might be aspects I’m overlooking due to my current skill level, but based on my understanding, this is how I perceive it!!!

Issue #3427, opened 2024-04-15
Spring Data JPA generates incorrect JPQL query for sorted pagination request with UNION clause


Pedroloma:
After updating my app from 3.0.8 to 3.1.10, I get an exception for a JPL with UNION clauses, when the query is pageable and sorted.

I've created a GitHub repo where you can reproduce the issue: https://github.com/Pedroloma/JpaWithUnionAndPageableAndSortedApplication

I think, that this issue could be related to this other one: https://github.com/spring-projects/spring-data-jpa/issues/2969

This is the exception thrown and its stack trace:

```
Caused by: org.hibernate.query.sqm.ParsingException: line 1:79 mismatched input 'ascUNION' expecting {<EOF>, ',', '.', '[', '+', '-', '*', '/', '%', '||', ASC, BY, DAY, DESC, EPOCH, FETCH, HOUR, LIMIT, MINUTE, MONTH, NANOSECOND, NULLS, OFFSET, QUARTER, SECOND, WEEK, YEAR}
	at org.hibernate.query.hql.internal.StandardHqlTranslator$1.syntaxError(StandardHqlTranslator.java:46)
	at org.antlr.v4.runtime.ProxyErrorListener.syntaxError(ProxyErrorListener.java:41)
	at org.antlr.v4.runtime.Parser.notifyErrorListeners(Parser.java:543)
	at org.antlr.v4.runtime.DefaultErrorStrategy.reportInputMismatch(DefaultErrorStrategy.java:327)
	at org.antlr.v4.runtime.DefaultErrorStrategy.reportError(DefaultErrorStrategy.java:139)
	at org.hibernate.grammars.hql.HqlParser.statement(HqlParser.java:361)
	at org.hibernate.query.hql.internal.StandardHqlTranslator.parseHql(StandardHqlTranslator.java:127)
	at org.hibernate.query.hql.internal.StandardHqlTranslator.translate(StandardHqlTranslator.java:77)
	at org.hibernate.query.internal.QueryInterpretationCacheStandardImpl.createHqlInterpretation(QueryInterpretationCacheStandardImpl.java:165)
	at org.hibernate.query.internal.QueryInterpretationCacheStandardImpl.resolveHqlInterpretation(QueryInterpretationCacheStandardImpl.java:147)
	at org.hibernate.internal.AbstractSharedSessionContract.interpretHql(AbstractSharedSessionContract.java:741)
[...]

PR #3429, opened 2024-04-16
Fix order by rendering for queries containing UNION 

christophstrobl:
Make sure to append space after order by clause and fix alias detection for wrapped sub select.
Also make sure to ignore alias used in subselect so they do not conflict with root ones.

Closes: #3427

### DIFF
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
@@ -1070,8 +1070,11 @@ class HqlQueryTransformerTests {
 
         String source = "SELECT tb FROM Test tb WHERE (tb.type='A') UNION SELECT tb FROM Test tb WHERE (tb.type='B')";
 		String target = createQueryFor(source, Sort.by("Type").ascending());
+		
+		System.out.println("target: " + target);
 
-		assertThat(target).contains(" UNION SELECT ").doesNotContainPattern(Pattern.compile(".*\\SUNION"));
+		assertThat(target).isEqualTo("SELECT tb FROM Test tb WHERE (tb.type = 'A') UNION SELECT tb FROM Test tb WHERE (tb.type = 'B') order by tb.Type asc");
+//		assertThat(target).contains(" UNION SELECT ").doesNotContainPattern(Pattern.compile(".*\\SUNION"));
 	}
 
 	@ParameterizedTest // GH-3427
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
@@ -17,6 +17,8 @@ package org.springframework.data.jpa.repository.query;
 
 import static org.assertj.core.api.Assertions.*;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import org.assertj.core.api.SoftAssertions;
@@ -24,11 +26,13 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.springframework.dao.InvalidDataAccessApiUsageException;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.jpa.domain.JpaSort;
 import org.springframework.lang.Nullable;
+import org.springframework.util.StringUtils;
 
 /**
  * Verify that HQL queries are properly transformed through the {@link JpaQueryEnhancer} and the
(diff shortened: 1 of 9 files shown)


spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
```diff
@@ -1043,6 +1047,40 @@ void createsCountQueryUsingAliasCorrectly() {
 		assertCountQuery("select distinct a, count(b) as c from Employee GROUP BY n","select count(distinct a, count(b)) from Employee AS __ GROUP BY n");
 	}
 
+	@Test // GH-3427
+	void sortShouldBeAppendedWithSpacingInCaseOfSetOperator() {
+
+        String source = "SELECT tb FROM Test tb WHERE (tb.type='A') UNION SELECT tb FROM Test tb WHERE (tb.type='B')";
+		String target = createQueryFor(source, Sort.by("Type").ascending());
```
schauder, line 1054: After this line `target` has the value 
```sql
SELECT tb 
    FROM Test tb 
    WHERE (tb.type = 'A') 
    order by tb.Type asc 
UNION 
SELECT tb 
    FROM Test tb 
    WHERE (tb.type = 'B') 
    order by tb.Type asc`
```
i.e. the `order by` is applied twice which is illegal, since according to answers here https://stackoverflow.com/questions/4715820/how-to-order-by-with-union-in-sql the later `order by` applies to the full `union` and an `order by` on the individual selects might be even illegal (probably depends on the database)

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
```diff
@@ -1043,6 +1047,40 @@ void createsCountQueryUsingAliasCorrectly() {
...
+
+		assertThat(target).contains(" UNION SELECT ").doesNotContainPattern(Pattern.compile(".*\\SUNION"));
```
schauder, line 1056: Since we put in an exact SQL statement we should be able to assert on a precise output and not rely on pattern matching and simiar.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
```diff
@@ -1043,6 +1047,40 @@ void createsCountQueryUsingAliasCorrectly() {
...
+	}
+
+	@ParameterizedTest // GH-3427
+	@ValueSource(strings = {"", "res"})
```
schauder, line 1060: A comment would be nice explaining why we need to run this with different prefixes.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
```diff
@@ -1043,6 +1047,40 @@ void createsCountQueryUsingAliasCorrectly() {
...
+	void sortShouldBeAppendedToSubSelectWithSetOperatorInSubselect(String alias) {
+
+		String prefix = StringUtils.hasText(alias) ? (alias + ".") : "";
```
schauder, line 1063: Again I think we should be able to and prefer a simple equals-assertion.

schauder (changes requested):

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
christophstrobl: since we do not know the target db nor how it handles order by we cannot imply ordering the last is the correct behaviour, nor do we have the means to define which of the selects to decorate. an alternative would be to raise an error and demand wrapping the  entire thing as it's done in one of the follow up tests.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
christophstrobl: seems legit

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
christophstrobl: deemed to be obvious - apparently isn't

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
christophstrobl: I'm not a fan of duplication.

christophstrobl:
we should additionally check the count query creation for `UNION`.

christophstrobl:
@schauder @mp911de - care to have another look at this one. 
Changed the sort transformation so that the provided `Sort` is now appended after the last query expression.

spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/HqlQueryTransformerTests.java
mp911de: Probably, let's wait and see what kind of reports we get. We can then still react to the feedback we receive.

christophstrobl: reading through IBM an Oracle SQL reference docs @schauder has a good point in the above mentioned query not being compliant (at least not in the form it is currently rendered)

> For example, the following is not valid (SQLSTATE 428FJ):
>`SELECT * FROM T1  `
>`ORDER BY C1       `
>`UNION             `
>`SELECT * FROM T2  `
>`ORDER BY C1       `
>
>
> The following example is valid:
> `(SELECT * FROM T1 `
> `ORDER BY C1)      `
> `UNION             `
> `(SELECT * FROM T2 `
> `ORDER BY C1)      `

mp911de:
Updated the HQL code to apply ordering after the last set operation, EQL did that already.

Issue #3410, opened 2024-03-26
Problem with Spring Data JPA with projections, specifications and SPEL


z8manu8z:
I'm having a problem using a bean in a SPEL `@Value` expression by combining Projections and Specifications.

Example project : [spring-jpa-projections-specs.zip](https://github.com/spring-projects/spring-data-jpa/files/14759947/spring-jpa-projections-specs.zip)

In the attached project, a projection returned by a `@Query` works correctly, while a projection returned by JpaSpecificationExecutor.findBy(Specification spec) returns an error during http serialization.

CityProjection : 

`@Value("#{@projectionHelper.fullName(target.name, target.country)}")
String getFullName();`

In CityServiceImpl line 79 : 

`CityProjection = this.cityRepository.findByNameAndCountry(name, country);` => works well

`CityProjection = this.cityRepository.findBy(new Specification<City>() {
			@Override
			public Predicate toPredicate(Root<City> root, CriteriaQuery<?> query, CriteriaBuilder criteriaBuilder) {
				return null;
			}
		}, q -> q.as(CityProjection.class).firstValue());`
		
=> fails with message : 

Resolved [org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: EL1057E: No bean resolver registered in the context to resolve access to bean 'projectionHelper']

How can I solve this ?

christophstrobl:
The fix unfortunately requires quite a few changes in the current implementation, so I'm afraid there's no easy work around for this at the moment.

z8manu8z:
@christophstrobl Ok thanks for your investigations !
I can adapt my implementation, no worry about that.
Can you show me how I can do ?

PR #3432, opened 2024-04-17
Fluent query API should use ProjectionFactory provided by the RepositoryFactory

christophstrobl:
This PR makes sure to push the `ProjectionFactory` down to the fluent query to make use of beans registered in the context. Prior to this change the fluent query variant would host its own factory not being aware of its surroundings.

Resolves: #3410

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadata.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadata.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.util.Optional;
 
 import org.springframework.data.jpa.repository.EntityGraph;
+import org.springframework.data.projection.ProjectionFactory;
 import org.springframework.lang.Nullable;
 
 /**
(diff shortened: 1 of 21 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadata.java
```diff
@@ -85,4 +86,11 @@ public interface CrudMethodMetadata {
 	 * @since 1.9
 	 */
 	Method getMethod();
+
```
mp911de, line 89: `JpaRepositoryImplementation` provides already customization options for repository implementations. What if we evolve that concept by moving `setRepositoryMethodMetadata` and `setEscapeCharacter` methods into a new interface, that would also accept `ProjectionFactory` and let both, `JpaRepositoryImplementation` and `QuerydslJpaPredicateExecutor` implement that interface?

christophstrobl: worth exploring - also in the light of having to deal with a similar arrangement in other store modules.

PR #3444, opened 2024-04-23
Fixed Unnecessary Computation Issue Caused by `String.format()`

dukbong:
This PR improved the code by using Assert methods instead of throwing IllegalArgumentException in if-else statements.

- 33eb67a : fixed an issue where the computation was processed immediately due to instant evaluation, regardless of the Assert outcome.

- [ ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
@@ -145,7 +145,7 @@ public class JpaQueryMethod extends QueryMethod {
 		this.annotationCache = new ConcurrentReferenceHashMap<>();
 
 		Assert.isTrue(!(isModifyingQuery() && getParameters().hasSpecialParameter()),
-				String.format("Modifying method must not contain %s", Parameters.TYPES));
+				() -> String.format("Modifying method must not contain %s", Parameters.TYPES));
 		assertParameterNamesInAnnotatedQuery();
 	}
 
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
@@ -347,11 +347,9 @@ public class JpaQueryMethod extends QueryMethod {
 
 		String query = getAnnotatedQuery();
 
-		if (query != null) {
-			return query;
-		}
-
-		throw new IllegalStateException(String.format("No annotated query found for query method %s", getName()));
+		Assert.notNull(query, () -> String.format("No annotated query found for query method %s", getName()));
+		
+		return query;
 	}
 
 	/**
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadataPostProcessor.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadataPostProcessor.java
@@ -119,10 +119,9 @@ class CrudMethodMetadataPostProcessor implements RepositoryProxyPostProcessor, B
 
 			MethodInvocation mi = currentInvocation.get();
 
-			if (mi == null)
-				throw new IllegalStateException(
-						"No MethodInvocation found: Check that an AOP invocation is in progress, and that the "
+			Assert.notNull(mi, "No MethodInvocation found: Check that an AOP invocation is in progress, and that the "
 								+ "CrudMethodMetadataPopulatingMethodInterceptor is upfront in the interceptor chain.");
+
 			return mi;
 		}
 
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/DefaultJpaContext.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/DefaultJpaContext.java
@@ -61,17 +61,12 @@ public class DefaultJpaContext implements JpaContext {
 
 		Assert.notNull(type, "Type must not be null");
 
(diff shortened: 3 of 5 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryMethod.java
```diff
@@ -347,11 +347,9 @@ public String getRequiredAnnotatedQuery() throws IllegalStateException {
 
 		String query = getAnnotatedQuery();
 
-		if (query != null) {
-			return query;
-		}
-
-		throw new IllegalStateException(String.format("No annotated query found for query method %s", getName()));
+		Assert.notNull(query, () -> String.format("No annotated query found for query method %s", getName()));
```
mp911de, line 350: This change changes behavior into throwing `IllegalArgumentException` instead of `IllegalStateException`. At that point, throwing an explicit `IllegalStateException` is fine so we should not change this method.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/DefaultJpaContext.java
```diff
@@ -61,17 +61,12 @@ public EntityManager getEntityManagerByManagedType(Class<?> type) {
...
 
 		List<EntityManager> candidates = this.entityManagers.get(type);
 
-		if (candidates.size() == 1) {
-			return candidates.get(0);
-		}
-
-		throw new IllegalArgumentException(
-				String.format("%s managed by more than one EntityManagers: %s", type.getName(), candidates));
+		Assert.isTrue(candidates.size() == 1, () -> String.format("%s managed by more than one EntityManagers: %s", type.getName(), candidates));
```
mp911de, line 68: One thought for here: The control flow is being replaced with calls to `Assert`. While assertions are fine for method-entry checks, usage here is kind-of unexpected. Also, we're always creating a capturing lambda leading to a bit more GC pressure.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JpaMetamodelEntityInformation.java
```diff
@@ -299,9 +296,7 @@ public Class<?> getType() {
 			// lazy initialization of idType field with tolerable benign data-race
 			this.idType = tryExtractIdTypeWithFallbackToIdTypeLookup();
 
-			if (this.idType == null) {
-				throw new IllegalStateException("Cannot resolve Id type from " + type);
-			}
+			Assert.notNull(this.idType, () -> String.format("Cannot resolve Id type from %s", type));
```
mp911de, line 299: This change changes `IllegalStateException` to `IllegalArgumentException`

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/CrudMethodMetadataPostProcessor.java
```diff
@@ -119,10 +119,9 @@ static MethodInvocation currentInvocation() throws IllegalStateException {
 
 			MethodInvocation mi = currentInvocation.get();
 
-			if (mi == null)
-				throw new IllegalStateException(
-						"No MethodInvocation found: Check that an AOP invocation is in progress, and that the "
+			Assert.notNull(mi, "No MethodInvocation found: Check that an AOP invocation is in progress, and that the "
```
mp911de, line 122: This change changes behavior into throwing `IllegalArgumentException` instead of `IllegalStateException`.

mp911de (changes requested):
Excessive switching from `if` statements that throw an exception into `Assert` utility usage is counterintuitive in many places. There are also occurrences that change the thrown exception type.

It makes sense to revisit `Assert` usages in combination with `String.format` to defer message creation. However, in places that explicitly throw exceptions, we would like to keep the code as-is.
Once an exception is being thrown, we need to construct the message and that is a fine arrangement already.

dukbong:
> Excessive switching from `if` statements that throw an exception into `Assert` utility usage is counterintuitive in many places. There are also occurrences that change the thrown exception type.
> 
> It makes sense to revisit `Assert` usages in combination with `String.format` to defer message creation. However, in places that explicitly throw exceptions, we would like to keep the code as-is. Once an exception is being thrown, we need to construct the message and that is a fine arrangement already.

I apologize for negatively impacting the open-source project management.
I will close this PR and resubmit it with only the necessary operations from commit 33eb67a.

Before sending the PR, I will ensure that it is thoroughly checked multiple times.

I'm sorry again for the inconvenience caused.

dukbong:
> > Excessive switching from `if` statements that throw an exception into `Assert` utility usage is counterintuitive in many places. There are also occurrences that change the thrown exception type.
> > It makes sense to revisit `Assert` usages in combination with `String.format` to defer message creation. However, in places that explicitly throw exceptions, we would like to keep the code as-is. Once an exception is being thrown, we need to construct the message and that is a fine arrangement already.
> 
> I apologize for negatively impacting the open-source project management. I will close this PR and resubmit it with only the necessary operations from commit [33eb67a](https://github.com/spring-projects/spring-data-jpa/commit/33eb67a11dbc5eee6dae10636595e3f9c635071e).
> 
> Before sending the PR, I will ensure that it is thoroughly checked multiple times.
> 
> I'm sorry again for the inconvenience caused.

#3445

mp911de:
First of all, do not worry, and there's no need to apologize. We're happy to provide guidance.

Please also note that you can force-push into a pull request until the final merge so that you can reuse already open pull requests.

dukbong:
Oh, I wasn't aware of that. I have now created a separate PR for the parts that urgently need to be fixed. I'll keep this in mind and use it more effectively going forward.

PR #3448, opened 2024-04-26
Introduce `Specification.getParameters()`

quaff:
Add the ability to set parameter values for Specification

Fix GH-3192


quaff:
> Closing as per [#3192 (comment)](https://github.com/spring-projects/spring-data-jpa/pull/3192#issuecomment-1797953668)

>> We discussed the topic as team and concluded that separation of the query and its parameters isn't a great programming model.
>>
>> For safe SQL query execution, the query declaration and the parameters are typically split (SELECT foo FROM bar where baz = :baz) as that is the mode which queries are build and parameters are passed. In arrangements where queries are colocated with parameters for execution (e.g. jdbcTemplate.execute("SELECT … where baz = :baz", myBaz)), you can immediately realize the relationship and parameter requirement.
>> We've been advocating a captive programming model for specifications in the following form:
>>
>> ```java
>> var spec = userHasFirstname("Foo").or(userHasLastname("Bar")).or(userHasLastname("Baz"));
>> repository.findAll(spec);
>> ```

I totally agree that, this PR colocate query and its parameters in single `Specification`, please see added tests.

mp911de:
We declined passing on parameters in an earlier pull request #3192. Parameters require a lot more design and a well-thought mechanism to avoid clashes across multiple parameters. Especially `allOf`, `and` are candidates where named parameters can override each other ending up with an invalid query state.

Currently, we do not have the bandwidth to work on the parameters design. We have this request however on our radar.

quaff:
> Especially `allOf`, `and` are candidates where named parameters can override each other ending up with an invalid query state.

`SpecificationComposition` will deny ambiguous parameter.

Issue #3286, opened 2023-12-30
`QueryTypeMismatchException` if the DTO has a no-args constructor


jhs512:
# PostRepository.java
```java
package com.ll.medium.domain.post.post.repository;

import com.ll.medium.domain.member.member.entity.Member;
import com.ll.medium.domain.post.post.entity.Post;
import org.springframework.data.domain.Page;
import org.springframework.data.domain.Pageable;
import org.springframework.data.jpa.repository.JpaRepository;

import java.util.List;
import java.util.Optional;

public interface PostRepository extends JpaRepository<Post, Long> {
    <T> Optional<T> findTop1ByOrderByIdDesc(Class<T> type);
}
```

# Post.java
```java
package com.ll.medium.domain.post.post.entity;

import com.ll.medium.domain.member.member.entity.Member;
import com.ll.medium.global.jpa.BaseTime.BaseTime;
import jakarta.persistence.Entity;
import jakarta.persistence.ManyToOne;
import lombok.*;

@Entity
@Builder
@NoArgsConstructor(access = AccessLevel.PROTECTED)
@AllArgsConstructor(access = AccessLevel.PROTECTED)
@Getter
@Setter
@EntityListeners(AuditingEntityListener.class)
@EqualsAndHashCode(onlyExplicitlyIncluded = true)
public class Post extends BaseTime {
    @Id
    @GeneratedValue(strategy = IDENTITY)
    @EqualsAndHashCode.Include
    private Long id;
    @CreatedDate
    private LocalDateTime createDate;
    @LastModifiedDate
    private LocalDateTime modifyDate;
    @ManyToOne
    private Member author;
    private String title;
    private String body;
    private boolean published;
}
```

# PostDtoImpl.java
```java
@Getter
@Setter
@NoArgsConstructor // When this annotation is present, the following error occurs
@AllArgsConstructor
public class PostDtoImpl {
    private long id;
    private LocalDateTime createDate;
    private LocalDateTime modifyDate;
    private long authorId;
    private String authorUsername;
    private String title;
    private String body;
    private boolean published;
}
```

# Error Msg
```
org.hibernate.query.QueryTypeMismatchException: Specified result type [com.ll.medium.domain.post.post.dto.PostDto] did not match Query selection type [com.ll.medium.domain.post.post.entity.Post] - multiple selections: use Tuple or array
[...]

PR #3449, opened 2024-04-26
Update Projection section in reference documentation.

christophstrobl:
Related to:  #3286

### DIFF
--- a/src/main/antora/modules/ROOT/pages/repositories/projections.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/projections.adoc
@@ -4,3 +4,28 @@
 include::{commons}@data-commons::page$repositories/projections.adoc[leveloffset=+1]
 
 NOTE: It is important to note that <<projections.dtos,Class-based projections>> with JPQL is limited to *constructor expressions* in your JPQL expression, e.g. `SELECT new com.example.NamesOnly(u.firstname, u.lastname) from User u`. (Note the usage of a FQDN for the DTO type!) This JPQL expression can be used in `@Query` annotations as well where you define any named queries. And it's important to point out that class-based projections do not work with native queries AT ALL. As a workaround you may use named queries with `ResultSetMapping` or the Hibernate specific https://docs.jboss.org/hibernate/orm/6.0/javadocs/org/hibernate/transform/ResultTransformer.html[`ResultTransformer`]
+
+[NOTE]
+====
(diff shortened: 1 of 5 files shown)


src/main/antora/modules/ROOT/pages/repositories/projections.adoc
```diff
@@ -4,3 +4,28 @@
 include::{commons}@data-commons::page$repositories/projections.adoc[leveloffset=+1]
 
 NOTE: It is important to note that <<projections.dtos,Class-based projections>> with JPQL is limited to *constructor expressions* in your JPQL expression, e.g. `SELECT new com.example.NamesOnly(u.firstname, u.lastname) from User u`. (Note the usage of a FQDN for the DTO type!) This JPQL expression can be used in `@Query` annotations as well where you define any named queries. And it's important to point out that class-based projections do not work with native queries AT ALL. As a workaround you may use named queries with `ResultSetMapping` or the Hibernate specific https://docs.jboss.org/hibernate/orm/6.0/javadocs/org/hibernate/transform/ResultTransformer.html[`ResultTransformer`]
+
+[NOTE]
+====
+Some JPA providers may require additional hints when working with <<projections.dtos,Class-based projections>> especially when using those directly with method derived queries.
```
schauder, line 10: That confuses me. `@PersistenceCreator` is a Spring Data annotation, I'm sure that JPA providers don't care about that. 

I would expect that the situation is more: We need either a constructor taking all the properties as arguments and if it isn't the only constructor we need it to be marked with `@PersistenceCreator`.

src/main/antora/modules/ROOT/pages/repositories/projections.adoc
```diff
@@ -4,3 +4,28 @@
...
+If you are facing errors like `JpaSystemException: Specified result type did not match Query selection type` make sure to provide a constructor hint via `@PersistenceCreator` to be passed on, as outlined below:
```
schauder, line 11: It isn't so much _outlined_ below but demonstrated, isn't it?

schauder (changes requested):

src/main/antora/modules/ROOT/pages/repositories/projections.adoc
mp911de: It's our code that determines input properties. We use our mechanism to discover whether there is a preferred constructor and the properties it takes.

PR #3452, opened 2024-04-28
Polishing

quaff:
1. `Specification` should be annotated with `@FunctionalInterface`
2. `SpecificationUnitTests` shouldn't be `Serializable`


mp911de:
Putting this on hold as we need to overhaul the `Specification` infrastructure before adding more changes to it.

mp911de:

Issue #3242, opened 2023-11-24
Redundant parameter in the Limit query


sergey-morenets:
Hi

Spring Data JPA introduced new Limit functionality to provide dynamic limiting. So if earlier we wrote the following query:

```
	List<Product> findTop3By();
	
	List<Product> findFirst3By();
```

and now we can use new approach:

`	List<Product> findBy(Limit limit);`

But I noticed that resulting SQL in the first two queries was :

`SELECT * from Product p1_0 fetch first ? rows only`

And now it's:

`SELECT * from Product p1_0 offset ? rows fetch first ? rows only`

So why do we need additional parameter offset? I guess it's redundant and not needed here.

mp911de:
This is because we convert by default `Limit` into `Pageable(0, limit)` and set the `Query.setFirstResult(…)` value implicitly.

Actually, Hibernate could omit that value as not setting the offset and setting offset zero should yield the same result.

quaff:
I will try to fix it at hibernate side.

quaff:
I've created https://github.com/hibernate/hibernate-orm/pull/7579 to fix this.

sergey-morenets:
> I've created [hibernate/hibernate-orm#7579](https://github.com/hibernate/hibernate-orm/pull/7579) to fix this.

Thank you, @quaff

quaff:
> > I've created [hibernate/hibernate-orm#7579](https://github.com/hibernate/hibernate-orm/pull/7579) to fix this.
> 
> Thank you, @quaff

FYI, Hibernate team decline that.

PR #3454, opened 2024-04-28
Ignore offset for `Limit` based query

quaff:
Avoid calling `Query.setFirstResult(0)` which maybe generated sql contains unwanted `offset 0`.

Fix GH-3242


quaff:
It's not likely that Hibernate will merge proposed PR linked to https://hibernate.atlassian.net/browse/HHH-17475, I think it's better to adapt at Spring Data JPA side.

mp911de:

PR #3456, opened 2024-04-29
Use provider built-in result count to reuse query if possible

quaff:
Hibernate introduce `SelectionQuery::getResultCount` since 6.5.0, It may be JPA Query method in the future.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
@@ -62,6 +62,7 @@ import org.springframework.util.Assert;
  * @author Сергей Цыпанов
  * @author Wonchul Heo
  * @author Julia Lee
+ * @author Yanming Zhou
  */
 public abstract class AbstractJpaQuery implements RepositoryQuery {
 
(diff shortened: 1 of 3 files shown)


christophstrobl:
Thank you @quaff for bringing this up. Did you eventually compare the resulting database interaction triggered by `getResultCount` to the `count` as it is done right now? Before we change the current behaviour I'd like to understand potential implications on both what the count query looks like as well as potential runtime implications.

quaff:
> Thank you @quaff for bringing this up. Did you eventually compare the resulting database interaction triggered by `getResultCount` to the `count` as it is done right now? Before we change the current behaviour I'd like to understand potential implications on both what the count query looks like as well as potential runtime implications.

Please refer to https://github.com/hibernate/hibernate-orm/blob/8c3b68cd784076a06a4bc2a7a3d67fc32904dba3/hibernate-core/src/main/java/org/hibernate/query/sqm/tree/select/SqmSelectStatement.java#L465

mp911de:
This PR isn't complete. I took a look at a complete implementation and pushed my variant on https://github.com/spring-projects/spring-data-jpa/tree/issue/3456.

mp911de:
Using `SelectionQuery.getResultCount()` only works when there's an enclosing transaction, otherwise the underlying `Session` will be closed.

Issue #3451, opened 2024-04-26
JPQL and HQL parser issue with class projection when FQDN includes "exp"


jgooley-sh:
Version: Latest 3.1.x

It appears that when using class based projection we can't have `exp` in our package name.

```
select new com.company.exp.thing.stuff.ClassName(
  e.id,
  cast(round(coalesce(SUM(e.amount), 0),2) as decimal) as amountTotal
)
from Experience e
group by e.id
```

Antlr output:
> line 1:22 mismatched input '.' expecting '('
> line 1:32 mismatched input '.' expecting {<EOF>, ',', EXCEPT, FROM, GROUP, INTERSECT, ORDER, UNION, WHERE}

christophstrobl:
@jgooley-sh thanks for reporting - we'll look into this.

PR #3457, opened 2024-04-29
Fix parsing entity names in JPQL query with package names that contain reserved words

christophstrobl:
Resolves: #3451

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa-parent</artifactId>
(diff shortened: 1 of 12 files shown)


jgooley-sh:
Not sure if this PR handles the reserved words anywhere in the query, but I found that the table alias couldn't be the reserved word either.

```
select exp
from Experience exp
```

mp911de:
> Not sure if this PR handles the reserved words anywhere in the query, but I found that the table alias couldn't be the reserved word either.

It doesn't allow reserved word usage for aliases. Isn't the purpose of reserved words that you cannot use these in arbitrary places?

PR #3463, opened 2024-05-07
Issue/3460

thjanssen:
Fix handling of position INOUT parameters when extracting output parameters (issue 3460)

Added position information to `ProcedureParameter` class, set it in `StoredProcedureAttributeSource.extractOutputParametersFrom(...)`, and use that position in `StoredProcedureJpaQuery.extractOutputParameterValue` to get the result value.
This change is based on the assumption that `NamedStoredProcedureQuery.parameters()` returns the parameters in the correct order. As far as I understand the rest of the code, it already makes this assumption when using positional parameters.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/StoredProcedureAttributes.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/StoredProcedureAttributes.java
@@ -88,7 +88,7 @@ class StoredProcedureAttributes {
 
 	private ProcedureParameter getParameterWithCompletedName(ProcedureParameter parameter, int i) {
 
-		return new ProcedureParameter(completeOutputParameterName(i, parameter.getName()), i+1, parameter.getMode(),
+		return new ProcedureParameter(completeOutputParameterName(i, parameter.getName()), parameter.getPosition(), parameter.getMode(),
 				parameter.getType());
 	}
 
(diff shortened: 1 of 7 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/StoredProcedureAttributes.java
```diff
@@ -87,7 +88,7 @@ private List<ProcedureParameter> getParametersWithCompletedNames(List<ProcedureP
 
 	private ProcedureParameter getParameterWithCompletedName(ProcedureParameter parameter, int i) {
 
-		return new ProcedureParameter(completeOutputParameterName(i, parameter.getName()), parameter.getMode(),
+		return new ProcedureParameter(completeOutputParameterName(i, parameter.getName()), i+1, parameter.getMode(),
```
christophstrobl, line 91: looks like we could reuse `parameter.getPosition()` here instead of `i+1`.

thjanssen: Yes, that should be better. I've added the change to this PR

thjanssen:
When this is fixed, do you think you could backport it to version 3.2.x?

christophstrobl:
I intend to pack port the fix if things work out as expected.

christophstrobl:
Thank you @thjanssen - merged to _main_ development line and back ported to _3.2.x_.

PR #3465, opened 2024-05-08
Polishing `SimpleJpaRepository`

quaff:
`TypedQuery.set*()` returns the same query instance


christophstrobl:
overlaps partially with #3468

Issue #3036, opened 2023-06-22
Specification toPredicate usage is violated in SimpleJpaRepository::delete


aryu-ki:
```java
@Override
	public long delete(Specification<T> spec) {

		CriteriaBuilder builder = this.em.getCriteriaBuilder();
		CriteriaDelete<T> delete = builder.createCriteriaDelete(getDomainClass());

		Predicate predicate = spec.toPredicate(delete.from(getDomainClass()), null, builder);

		if (predicate != null) {
			delete.where(predicate);
		}

		return this.em.createQuery(delete).executeUpdate();
	}
```
This is the definition of delete method in SimpleJpaRepository. Clearly, query param sent is null, but in Specification::toPredicate we have this doc:
```java
/**
	 * Creates a WHERE clause for a query of the referenced entity in form of a {@link Predicate} for the given
	 * {@link Root} and {@link CriteriaQuery}.
	 *
	 * @param root must not be {@literal null}.
	 * @param query must not be {@literal null}.
	 * @param criteriaBuilder must not be {@literal null}.
	 * @return a {@link Predicate}, may be {@literal null}.
	 */
	@Nullable
	Predicate toPredicate(Root<T> root, CriteriaQuery<?> query, CriteriaBuilder criteriaBuilder);
```

aryu-ki:
@mp911de You should change documentation string of toPredicate if I'm right, because it's a confusing doc. I could find the issue fast, but in my code, there were no obvious signs of what was going on. Other developers might get confused as well. Who knows how much time is it gonna cost to someone

ghostg00:
I also think that toPredicate's method documentation should be changed, It does cause confusion

PR #3468, opened 2024-05-10
Fix passing 'null' argument to parameter annotated as @NotNull

quaff:
Fix GH-3036


christophstrobl:
I think we should update the documentation and nullability contract annotations so they document what is currently allowed.

gbrehmer:
this would solve the current Kotlin incompatibility as well

christophstrobl:
thanks for the heads up - moved this one up in the list.

christophstrobl:
Thank you @quaff !

PR #3471, opened 2024-05-11
Add convenient method `JpaRepository::clear` as companion of `flush`

quaff:
Sometimes we need to clear persistence context, especially for writing unit tests. We need to inject `EntityManager` before this commit, it is a bit inconvenient.


mp911de:
We do not want to mirror `EntityManager` API on the repository. Exposing the `flush` method already clutters the repository API and we do not want to continue in that way. We suggest implementing a repository fragment for your convenience.

Issue #3462, opened 2024-05-07
Add `camelCase` to `snake_case` fallback for native query projections


cjacky475:
Hello, I have my **Friend** repository class from **friend** Modulith module: 
 
```
@Repository
interface FriendRepository: JpaRepository<Friend, Long> {
```

For testing purposes I will create simple function to get users within Friend repository:

```
 @Query(value = "SELECT * FROM users WHERE users.id = 558978864945139712 ", nativeQuery = true)
 fun getUsers(): Optional<UserProjection>
```
The interface for the User entity is:

```
interface UserProjection {
    fun getId(): Long
    fun getUsername(): String
    fun getEmail(): String
    fun getFullName(): String
}
```
The User entity has these fields (others emitted):

```
val email: String = "",
var fullName: String = "",
```

However, the REST API returns null on **fullName** field. 

How does that work exactly? Why my projection fails for camel case fields when projection class (in this instance **UserProjection** is in different module?

**--------------------- EDIT ---------------------**

Okay, so I finally understood what was happening. I had to create two projection classes, one for functions that use native query and one for those who do not. 

For functions that use native query, interface projection for camel case fields must be specified as:
```
 @JsonProperty("fullName")
 fun getFull_name(): String
```

for projection that do not use native queries, I can use:

```
fun getFullName(): String
```

I still do not understand why is that. Is there a more appropriate solution for using only single projection so it would satisfy both native and non native queries? Cheers.

quaff:
>> By default, Spring Boot configures the physical naming strategy with CamelCaseToUnderscoresNamingStrategy.

The column name is `full_name`,  you could use `select full_name as fullName` as workaround or switch to another NamingStrategy like `org.hibernate.boot.model.naming.PhysicalNamingStrategyStandardImpl`.

cjacky475:
Hi, @quaff, thanks for the answer. 

> The column name is full_name, you could use select full_name as fullName as workaround

This would become very inefficient and unmaintainable to write out all individual fields for selection, wouldn't it? `SELECT id, full_name as fullName, etc...`

> switch to another NamingStrategy like org.hibernate.boot.model.naming.PhysicalNamingStrategyStandardImpl

But this is not going to change the fact that when using native and non-native queries the return value is interpreted differently? I use PostgreSQL and return values are snake like full_name, also when submitting they have to be snake case.

Cheers.

quaff:
[...]

PR #3472, opened 2024-05-13
Support converting underscores column name to camel case property name for projecting native query

quaff:
> By default, Spring Boot configures the physical naming strategy with CamelCaseToUnderscoresNamingStrategy.

Then we should convert underscores back to camel case.

Fix GH-3462

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
@@ -47,6 +47,7 @@ import org.springframework.data.repository.query.RepositoryQuery;
 import org.springframework.data.repository.query.ResultProcessor;
 import org.springframework.data.repository.query.ReturnedType;
 import org.springframework.data.util.Lazy;
+import org.springframework.jdbc.support.JdbcUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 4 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractJpaQuery.java
```diff
@@ -378,6 +403,14 @@ public boolean containsKey(Object key) {
 					tuple.get((String) key);
 					return true;
 				} catch (IllegalArgumentException e) {
+					if (nativeQuery) {
+						try {
+							tuple.get(JdbcUtils.convertPropertyNameToUnderscoreName((String) key));
```
mp911de, line 408: Instead of customizing the `Map`, it would make sense to have a custom `Tuple` wrapper that handles the conversion. 

Consider a tuple that would contain `first_name` and `firstName` elements. It is impossible to tell in which order these should appear in a map nor which value to return in a `Set<Map.Entry>`.

Attempting snake-case to camel-case conversion on value absence makes sense, for the rest we should keep the functionality as-is.

The `Tuple` wrapper would delegate all enumerative calls to the `Tuple` delegate while `get(…)` calls are augmented for property name conversion.

quaff: Good point, updated now, please review again.

mp911de:

PR #3491, opened 2024-05-30
Use List.of instead of Arrays.asList In the TO_REPLACE field.

Lee-ChungMu:
Use List.of instead of Arrays.asList In the TO_REPLACE field.

The Arrays.asList method can change elements. SEARCH_OUTPUT_FIELDS is an immutable field, so it has been improved to util List.of.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EscapeCharacter.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EscapeCharacter.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.data.jpa.repository.query;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EscapeCharacter.java
```diff
@@ -31,7 +30,7 @@
 public final class EscapeCharacter {
 
 	public static final EscapeCharacter DEFAULT = EscapeCharacter.of('\\');
-	private static final List<String> TO_REPLACE = Arrays.asList("_", "%");
+	private static final List<String> TO_REPLACE = List.of("_", "%");
```
mp911de, line 33: We have many more occurences where we could use `List.of(…)` or `new HashSet<>(Arrays.asList(…))`. Care to have a look at these as well? Merging just one line change without considering the other places takes us more time than value it creates.

mp911de (changes requested):

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EscapeCharacter.java
Lee-ChungMu: I really didn’t think it through. Thank you for giving me good advice.

Issue #3496, opened 2024-06-05
BadJpqlGrammarException when parsing query containing lateral keyword used as property name.


christophstrobl:
Similar issue as with the type keyword (see: #3056).
While the _JPQL_ variant seems to be fine with it the _HQL_ one barks at the keyword.
Nevertheless we should add tests for all parsers.

mp911de:
Do we run into a similar issue with EQL/JPQL?

PR #3497, opened 2024-06-05
Fix query parsing error for properties using reserved lateral keyword.

christophstrobl:
Add tests for existing parsers and fix the HQL one.
Should be back ported to _3.3.x_ and _3.2.x_.

Closes: #3496

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa-parent</artifactId>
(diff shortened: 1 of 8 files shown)


SergeVolodin:
@christophstrobl, Good Day.

Is there any possibility of porting this change to Spring Boot 3.1.*

Thanks.

christophstrobl:
@SergeVolodin `data-jpa:3.1.x` is unfortunately already out of [OSS maintenance](https://spring.io/projects/spring-data-jpa#support).

schauder:
Merged and ported back to 3.2

PR #3512, opened 2024-06-15
Add Detailed Pessimistic and Optimistic Locking Examples to Spring Data JPA

ibrahimbayburtlu:


christophstrobl:
Thank you for taking the time to provide this PR. While we do see the desire to have as much information as possible in one place, we do however not want to repeat what a specific `jakarta.persistence.LockModeType` is used for, but keep it on a level that explains the concept of using `@Lock` with query methods.

Issue #3476, opened 2024-05-15
Sorting of Pageable.unpaged(sort) is ignored by JPA Repository


azizairo:
I have case when i need unpaged result with sort order. So to implement this i just used Pageable.unpaged method that allows to pass Sort parameter and I am expacting that jpa repository will return unpaged but sorted result. But it is not working cause jpa repostiory ignores sorting for unpaged  request:

```java
@Override
	public Page<T> findAll(Pageable pageable) {

		if (pageable.isUnpaged()) {
			return new PageImpl<>(findAll());
		}

		return findAll((Specification<T>) null, pageable);
	}
```
Is it possible to change behavior to this:
```java
@Override
	public Page<T> findAll(Pageable pageable) {

		if (pageable.isUnpaged()) {
                        if(pageable.getSort().isSorted()) {
                              return new PageImpl<>(findAll(pageable.getSort()));
                        }
			return new PageImpl<>(findAll());
		}

		return findAll((Specification<T>) null, pageable);
	}
```

quaff:
I think it's reasonable, It's better to submit a PR instead.

azizairo:
Ok, I will try to work on PR

PR #3517, opened 2024-06-19
Modify getQuery(Specification, Pageable)

birariro:
#3476 

This problem is not just happening with ```findAll(pageable pageable)```
also occur in ```findAll(Specification<T> spec, Pageable pageable)```

How about fixing it like this?
```java
@Override
public Page<T> findAll(Pageable pageable) {
	return findAll((Specification<T>) null, pageable);
}
```

```java
protected TypedQuery<T> getQuery(@Nullable Specification<T> spec, Pageable pageable) {
	return getQuery(spec, getDomainClass(), pageable.getSort());
}
```


birariro:
[commit](https://github.com/spring-projects/spring-data-jpa/commit/38a11d04451b8335e183b0edb2e2e0895f7ff2a5) Close PR

Issue #1280, opened 2016-07-05
Consider `NULLS` precedence using `Sort`


spring-projects-issues:
**[Vladimir Domashnev](https://jira.spring.io/secure/ViewProfile.jspa?name=v.domashnev@gmail.com)** opened **[DATAJPA-925](https://jira.spring.io/browse/DATAJPA-925?redirect=false)** and commented

As of 1.10.2, `toJpaOrder(Order, Root<?>, CriteriaBuilder)` method of `org.springframework.data.jpa.repository.query.QueryUtils` ignores null precedence, that is provided by `org.springframework.data.domain.Sort.Order` nullHandling propery, which is present since 1.7.x. It is used by `org.springframework.data.jpa.repository.support.SimpleJpaRepository.getQuery(Specification...)`, thus affecting many methods of `org.springframework.data.jpa.repository.JpaRepository` interface

---

**Affects:** 1.10.2 (Hopper SR2)

**Reference URL:** https://github.com/spring-projects/spring-data-jpa/blob/master/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java

**Issue Links:**
- [DATAJPA-754](https://jira.spring.io/browse/DATAJPA-754) NullHandling.(NULLS_LAST and NULLS_FIRST)  does not work
 (_**"is duplicated by"**_)
- [DATAJPA-1302](https://jira.spring.io/browse/DATAJPA-1302) NullHandling.(NULLS_LAST and NULLS_FIRST) does not work
 (_**"is duplicated by"**_)
- [DATAJPA-1725](https://jira.spring.io/browse/DATAJPA-1725) Using nullsFirst() or nullsLast() on Sort and JpaSort has no effect on database query
 (_**"is duplicated by"**_)
- [DATACMNS-904](https://jira.spring.io/browse/DATACMNS-904) NullHandling not working
 (_**"is duplicated by"**_)
- [DATAJPA-825](https://jira.spring.io/browse/DATAJPA-825) Take into account org.springframework.data.domain.Sort.NullHandling for JPA Criteria

4 votes, 11 watchers

gregturn:
This is pending https://github.com/eclipse-ee4j/jpa-api/issues/76.

mp911de:
Finally, the JPA spec issue has been addressed so we can proceed here for the JPA 3.2 spec. Additionally, we should extend our parsers via #3108

gregturn:
[...]

PR #3526, opened 2024-06-27
Update query parsers with newly introduced JPA 3.2 keywords 

christophstrobl:
This PR updates the query parsers to deal with new keywords/functions in JPQL available via JPA 3.2.

* union
* intersect
* except
* cast
* left
* right
* replace
* ||

`NULL` precedence (#1280) for result ordering is not part of the changes and needs to be done in another PR.


schauder (approved):
I commented on a couple of places where some refactoring could improve the code. But there is no severe concern from my side.

mp911de:
Superseded by #3695

PR #3529, opened 2024-07-01
Support NULLS {FIRST | LAST} in JPQL queries

christophstrobl:
This PR adds support for parsing and appending order by items that define a `NULL` precedence.
Supporting `NULL` precedence for `Specifications` and the `Criteria` API needs to be done separately.

Related to: #1280

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa-parent</artifactId>
-	<version>3.4.0-SNAPSHOT</version>
+	<version>3.4.x-GH-1280-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data JPA Parent</name>
--- a/spring-data-envers/pom.xml
+++ b/spring-data-envers/pom.xml
@@ -5,12 +5,12 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-envers</artifactId>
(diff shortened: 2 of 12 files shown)


kctang:
Hi @mp911de ,

https://github.com/spring-projects/spring-data-jpa/blame/7f13a04c414ad210e30ef647fda75590e28526cb/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java#L756

I am getting "Applying Null Precedence using Criteria Queries is not yet supported." with 3.4.0.

Is this Null Precedence using Criteria Queries supported yet? If so, can we remove the conditional throw? 

This is kind of blocking my app's migration from Spring Boot 2.7 to Spring Boot 3.4.

Thanks for reading my message & hope to get a solution soon.

mp911de:
> Null Precedence using Criteria Queries

Null Precedence using Criteria Query is specified with JPA 3.2 (https://jakarta.ee/specifications/persistence/3.2/). Before that, we cannot set null precedence through the API.

kctang:
Since Spring Data JPA 4.0 will be based on JPA 3.2, why do we throw "unsupported operation exception" from Spring Data JPA 3.4 (as it only needs to support JPA 3.1)? "Null Precedence using Criteria Query" is not a specified feature in prior JPA versions.

What is the impact of not having this exception check in Spring Data JPA 3.4? If there's no runtime errors, I would recommend not having this conditional exception check.

From my tests, without exception check (in Spring Data v3.3), my app's repository queries seems to be working.

Issue #3440, opened 2024-04-22
Align EnableJpaRepositories with changes in data-commons.


christophstrobl:
See: spring-projects/spring-data-commons#3083

PR #3551, opened 2024-07-31
Add support for using custom BeanNameGenerator.

christophstrobl:
Closes: #3440


mp911de (approved):
Feel free to adopt the other modules as well.

PR #3552, opened 2024-07-31
Migrate build to Spring Develocity Conventions extension

erichaagdev:
This PR migrates the build to use Spring's Develocity Conventions Maven extension.

https://github.com/spring-io/develocity-conventions

With this, explicit configuration of the Develocity Maven extension is no longer required, as all configuration is encapsulated within the Spring extension.

Additionally, build cache authentication now uses the Develocity access key, so separate build cache credentials are no longer required. The `DEVELOCITY_ACCESS_KEY` environment variable is enough for both build scan publishing and remote build cache access.


mp911de:

PR #3564, opened 2024-08-07
Optimize entity deletion in SimpleJpaRepository

Seol-JY:
Optimize the `delete` method in `SimpleJpaRepository` to improve performance in certain scenarios. The main change is to first check if the entity is already managed by the EntityManager before performing any database operations.

- Reordered the deletion logic to prioritize managed entities
- Added a check for `entityManager.contains(entity)` at the beginning of the method
- Moved the database lookup (`entityManager.find()`) to only occur for non-managed entities

The goal of this change is to reduce unnecessary database lookups when deleting managed entities, potentially improving performance in high-throughput scenarios.

I think this optimization is particularly effective when using the `deleteById` method:

Please point out any potential risks or edge cases where this change might lead to unexpected behavior.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
@@ -93,6 +93,7 @@ import org.springframework.util.Assert;
  * @author Yanming Zhou
  * @author Ernst-Jan van der Laan
  * @author Diego Krupitza
+ * @author Seol-JY
  */
 @Repository
 @Transactional(readOnly = true)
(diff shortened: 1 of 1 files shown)


quaff:
> The goal of this change is to reduce unnecessary database lookups when deleting managed entities

`entityManager.find(type, entityInformation.getId(entity))` will not trigger database lookup If `entityManager.contains(entity)`.

Seol-JY:
> > The goal of this change is to reduce unnecessary database lookups when deleting managed entities
> 
> `entityManager.find(type, entityInformation.getId(entity))` will not trigger database lookup If `entityManager.contains(entity)`.

Thank you, @quaff,  I understand that this might mean the intended performance improvement isn't actually achieved.

Given this, would you say that the proposed change is unnecessary, or do you think there might still be benefits from a different perspective?

mp911de:

PR #3565, opened 2024-08-08
Fix typo in assertion message

Seol-JY:
Addresses a minor typographical error in the assertion message within the code. The message "Sample must not be null" has been corrected to "Example must not be null".


mp911de:

PR #3576, opened 2024-08-12
Cleaned up EnversRevisionRepositoryImpl

arefbehboudi:
Cleaned up EnversRevisionRepositoryImpl the code to enhance readability and maintainability


schauder:
Thanks for submitting this PR.

We try to avoid the stream API since it is prone to introduce performance issues. 
This is not a problem for 99.9% of the application, but it is a problem for libraries like Spring Data.

I merged the rest of your changes.

arefbehboudi:
Thank you for your time.  
Can you explain or recommend an article about Stream API performance issues?

schauder:
>Can you explain or recommend an article about Stream API performance issues?

No, sorry. This is based on our own experience. Years back we replaced a lot of stuff with streams, just to change it back after we realised it had a performance impact.

Again: For most applications this is not an issue, except for when you are on some really hot code path. But with a library almost anything might end up being part of a hot code path for someone.

arefbehboudi:
Thank you so much @schauder

PR #3580, opened 2024-08-14
Refactor type checks and improve code readability

Seol-JY:
- Replace isInstance() method calls with instanceof operator for improved readability
- Remove unnecessary type checks and castings in postProcessBeanFactory method

Some changes to `instanceof refactoring` are already covered here https://github.com/spring-projects/spring-data-jpa/pull/3465


mp911de:

Issue #3360, opened 2024-02-09
I want to do a countribution where methods with batch keywords are processed through the persistence context.


sami355-24:
While looking for query optimization, I learned about the Batch keyword provided by Data JPA. And I also learned that the performance varies depending on the presence or absence of the Batch keyword. However, when adding the Batch keyword, I found that the "IN" keyword sends the query at once, but it processes it without going through the persistence context. 

In response, I thought it would be good to modify the code so that methods with the Batch keyword are processed through the persistence context. If my idea feels valid, can I do a countribution on this?

mp911de:
Would you mind providing some pseudo-code that explains what you want to achieve? We currently have a difficult time understanding what you're up to.

sami355-24:
I'm not familiar with English, so I used a translator to leave an issue. So the writing can be a little awkward.

First of all, I'll tell you about the code that made me think of contribution with an example.
If you use deleteAll provided by First Data Jpa, delete the entity one at a time through the persistence context, so `@PostRemove`, which is called when the value is deleted in the persistence context, works fine.

```java
@SpringBootTest
class MemberRepositoryTest {

    @Autowired
    private MemberRepository memberRepository;

    @BeforeEach
    public void setUp() {
        for (int i = 0; i < 10; i++) {
            memberRepository.save(Member.builder()
                    .clubName("test" + i)
                    .build());
        }
    }

    @Test
    @DisplayName("PostRemove is invoked when deleting using deleteAll.")
    public void deleteAllTest() {
        memberRepository.deleteAll();
    }
}
```

result is..
<img width="1470" alt="image" src="https://github.com/spring-projects/spring-data-jpa/assets/72246411/f0bcd389-e2a8-409c-aaae-20f799de878e">

And if you delete it using the method with the Batch keyword provided by Data JPA, `@PostRemove` will not be called.
```java
@SpringBootTest
class MemberRepositoryTest {

    @Autowired
    private MemberRepository memberRepository;

    @BeforeEach
    public void setUp() {
        for (int i = 0; i < 10; i++) {
            memberRepository.save(Member.builder()
                    .clubName("test" + i)
                    .build());
        }
    }

    @Test
    @DisplayName("PostRemove is not invoked when deleting using DeleteAll.")
    public void deleteAllInBatchTest() {
        memberRepository.deleteAllInBatch();
    }

}
```

result is..
[...]

PR #3591, opened 2024-08-24
Enhance Batch Deletion Logic by Handling Entities and IDs in Persistence Context

sami355-24:
This contribution improves the batch deletion functionality in JPA repositories by ensuring that both entities and their IDs passed to deleteAllInBatch and deleteAllByIdInBatch methods are properly handled within the persistence context.

related: #3360 


mp911de:
Thanks for your pull request. Both `Batch` methods leverage `delete` queries to remove objects without loading them into memory (see also their Javadoc). This is an intentional design choice to perform removal as quickly as possible and without notifying listeners.

That being said, we want to keep the design as-is. Also, we cannot call `EntityManager.clear(…)` because that would remove all entities from the persistence context, potentially also these that require flushing.

PR #3592, opened 2024-08-25
Cleanup code

arefbehboudi:
- I added Assert.notNull to places where it was missing.
- Removed unnecessary throw Exception statements.
- I made the variables final, where they were only defined once.


mp911de:

PR #3595, opened 2024-08-26
feat: refactor AbstractStringBasedJpaQuery for replace statement lambda

mjimani:
Refactor AbstractStringBasedJpaQuery for replace statement lambda with expression lambda


mp911de:
Thanks for your contribution. We would like to keep the code as-is for improved debugabbility.

PR #3601, opened 2024-08-29
Fix `limit` of `FetchableFluentQueryBySpecification`.

### LINKED ISSUE #3600: `limit` in `FetchableFluentQueryBySpecification` addes duplicate `order by` clause
As this code in `FetchableFluentQueryBySpecification` shows, adding a `limit` applies the current sorting again.

https://github.com/spring-projects/spring-data-jpa/blob/b00220b50d789b1926f55df7fb7b3360d0486755/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryBySpecification.java#L96-L103

In my application, this is causing duplicate `order by` clause to be included in the SQL generated like `order by q1_0."id" desc,q1_0."id" desc`, as I call `.sortBy` first, and then `.limit`.

Of course, it's easy to fix on the application's side by calling `.limit` first, but I think this is a bug in the framework, as the other class `FetchableFluentQueryByPredicate` implementing the same interface does not have this behavior.

JingBh:
A very simple fix that closes #3600

This fix makes the two implementations of `FluentQuery.FetchableFluentQuery` behave the same (which they formerly don't), so I think it's not an API change.


mp911de:

Issue #2989, opened 2023-05-30
Provide an option to disable `JSqlParserQueryEnhancer`


onukristo:
When JSqlParser is in classpath, the `org.springframework.data.jpa.repository.query.QueryEnhancerFactory` enables `org.springframework.data.jpa.repository.query.JSqlParserQueryEnhancer`.

However JSqlParser is quite CPU and memory allocation expensive, and it creates a new thread every time a query gets parsed.
Also, while the `JSqlParser` is very useful for various use cases, it is not perfect, and can not parse all the queries. We have encountered even bugs, when parsing a more complex SQL, goes to endless loop.

One of our services reported 15% more CPU total, just from this component.

Essentially, we have JSqlParser in classpath for other reasons, but don't want all the negative effects from `JSqlParserQueryEnhancer`.

Currently we disable it via byte-code manipulation:
```java
new ByteBuddy()
          .redefine(clazz)
          .method(named("isJSqlParserInClassPath"))
          .intercept(FixedValue.value(false))
          .make()
          .load(clazz.getClassLoader(), ClassReloadingStrategy.fromInstalledAgent());
```

However, it is not maintainable solution as we would most likely need to revisit this every time `spring-data-jpa` gets upgraded.

Please provide another mechanism, preferably a Spring Boot property, to disable this mechanism.

mp911de:
Do you happen to have a profiler recording that you could attach to this ticket? I'm wondering whether we could generally have some caching to avoid parsing the query over and over again.

onukristo:
We have, but unfortunately I can not share any internals of our services. Corporate rules :)

But looking the profile, the caching would help greatly for sure and would fix the issue for us, I think.

For example we have a public library for measuring application side database access, via JSqlParser. We [use cache](https://github.com/transferwise/tw-entrypoints/blob/master/tw-entrypoints/src/main/java/com/transferwise/common/entrypoints/tableaccessstatistics/TableAccessStatisticsSpyqlListener.java#L76) there. 

I coincidentally just added a mechanism against the JSqlParser creating new threads and against those endless loops on rare queries as well: https://github.com/transferwise/tw-entrypoints/pull/31

This could also implemented together with the caching, just in case.

gregturn:
I'm wondering if there isn't a suitable way to "opt out" via a Spring Boot property setting that in turn could feed into Spring Data JPA?

[...]

PR #3623, opened 2024-09-26
Add configuration option to disable usage of JSqlParser for native queries.

christophstrobl:
This commit introduces the `spring.data.jpa.query.native.parser` property that allows to switch native query parsing to the default internal parser for scenarios where `JSqlParser` is on the classpath but should not be used by spring-data-jpa.

See: #2989

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
@@ -17,8 +17,11 @@ package org.springframework.data.jpa.repository.query;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.core.SpringProperties;
 import org.springframework.data.jpa.provider.PersistenceProvider;
+import org.springframework.lang.Nullable;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * Encapsulates different strategies for the creation of a {@link QueryEnhancer} from a {@link DeclaredQuery}.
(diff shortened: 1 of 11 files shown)


christophstrobl:
I do not consider the `default` option to be expressive enough to hint towards its behaviour, but I failed to come up with a better wording for it.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
```diff
@@ -82,4 +73,56 @@ public static QueryEnhancer forQuery(DeclaredQuery query) {
...
+
+		AUTO, DEFAULT, JSQL;
+
+		static final String NATIVE_PARSER_PROPERTY = "spring.data.jpa.query.native.parser";
+
+		private static NativeQueryEnhancer from(@Nullable String name) {
+			if (!StringUtils.hasText(name)) {
+				return AUTO;
+			}
+			return NativeQueryEnhancer.valueOf(name.toUpperCase());
```
mp911de, line 103: How about `ObjectUtils.caseInsensitiveValueOf(…)`?

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
```diff
@@ -82,4 +73,56 @@ public static QueryEnhancer forQuery(DeclaredQuery query) {
...
+		}
+
+		/**
+		 * @param classLoader ClassLoader to look up available libraries.
+		 * @return the current selection considering classpath avialability and user selection via
+		 *         {@link #NATIVE_PARSER_PROPERTY}.
+		 */
+		static NativeQueryEnhancer select(ClassLoader classLoader) {
+
+			if (!ClassUtils.isPresent("net.sf.jsqlparser.parser.JSqlParser", classLoader)) {
```
mp911de, line 113: For Graal Native, wouldn't we want to keep this check within a field?

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
```diff
@@ -17,29 +17,29 @@
...
+ * @author Christoph Strobl
  * @since 2.7.0
  */
 public final class QueryEnhancerFactory {
 
 	private static final Log LOG = LogFactory.getLog(QueryEnhancerFactory.class);
-
-	private static final boolean jSqlParserPresent = ClassUtils.isPresent("net.sf.jsqlparser.parser.JSqlParser",
-			QueryEnhancerFactory.class.getClassLoader());
+	private static final NativeQueryEnhancer NATIVE_QUERY_ENHANCER;
```
mp911de, line 38: With `SpringProperties` falling back to System Properties, wouldn't we want to keep the selection dynamic to allow changing system property values at runtime?

More with the rationale of setting the value later and not so much for flipping it.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
christophstrobl: the check would still work just not be initialized at build time which is not necessary since its done only once and forgotten afterwards.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryEnhancerFactory.java
christophstrobl: While we could do it, I'd rather not have this configurable. Allowing to change values at runtime might lead to interesting side effects in its best case, and we'd have to do the lookup every single time.

mp911de:

Issue #3636, opened 2024-10-07
Allow returning Map from the repository methods


mipo256:
I've not found any similar issue, so I'm creating a new one.

**Problem:**

The problem is that sometimes native queries are quite complicated, and it is possible that they would return unknown at declaration time number of columns. I'm not saying it is a great way of doing things, but that's just what we have in our case, for instance. 

And in this case we cannot really return any kind of projection, since we do not know anything about the structure of the result set being returned at compile time.

**Solution (Proposal)**:

I think it makes sense to consider allowing returning the `Map` from the repository query method, like:

```
@Query(value = "...", nativeQuery=true)
Map<String, Object> findSomethingComplex();
```

In this case, the key is supposed to be the column name, and the value is just a raw column value.

christophstrobl:
Thank you for getting in touch. 
Repository finder methods in data-jpa using `Map` or `List<Map>` as return type will give you something called a `TupleBackedMap` delegating calls to an underlying `jakarta.persistence.Tuple`. So actually the outlined code from above should already work as expected. 

I had a look at our reference documentation but could not spot anything mentioning this behaviour. So, unless I'm missing something here, this seems to be a documentation issue.

mipo256:
Hey @christophstrobl!
Would it also work for the methods where we expect a single result from the database? Like single `Map`, not a `List`.

christophstrobl:
Yes, you could have something like:
```
@Query(value = "SELECT * from book", nativeQuery=true)
List<Map<String, Object>> findSomething();

@Query(value = "SELECT * from book WHERE id = 1", nativeQuery=true)
Map<String, Object> findSomethingSingle();
```

I'll update the documentation.

PR #3637, opened 2024-10-08
Update native query documentation.

christophstrobl:
Resolves: #3636

### DIFF
--- a/src/main/antora/modules/ROOT/pages/jpa/query-methods.adoc
+++ b/src/main/antora/modules/ROOT/pages/jpa/query-methods.adoc
@@ -318,6 +318,26 @@ Valid values are (case-insensitive):
 
 A similar approach also works with named native queries, by adding the `.count` suffix to a copy of your query. You probably need to register a result set mapping for your count query, though.
 
+Next to obtaining mapped results, native queries allow you to read the raw `Tuple` from the database by choosing a `Map` container as the method's return type.
+The resulting map contains key/value pairs representing the actual database column name and the value.
+
+.Native query retuning raw column name/value pairs
+====
+[source, java]
+----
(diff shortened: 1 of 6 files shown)


src/main/antora/modules/ROOT/pages/jpa/query-methods.adoc
```diff
@@ -318,6 +318,26 @@ Valid values are (case-insensitive):
...
+  Map<String, Object> findRawMapByEmail(String emailAddress); <1>
+
+  @NativeQuery("SELECT * FROM USERS WHERE LASTNAME = ?1")
+  List<Map<String, Object>> findRawMapByLastname(String lastname); <2>
+}
+----
+<1> Single `Map` result backed by a `Tuple`.
+<2> Multiple `Map` results backed by ``Tuple``s.
+====
+
```
mp911de, line 340: We should mention that this only works for Hibernate.

src/main/antora/modules/ROOT/pages/jpa/query-methods.adoc
```diff
@@ -318,6 +318,26 @@ Valid values are (case-insensitive):
...
+The resulting map contains key/value pairs representing the actual database column name and the value.
+
+.Native query retuning raw column name/value pairs
+====
+[source, java]
+----
+public interface UserRepository extends JpaRepository<User, Long> {
+
+  @NativeQuery("SELECT * FROM USERS WHERE EMAIL_ADDRESS = ?1")
+  Map<String, Object> findRawMapByEmail(String emailAddress); <1>
```
mp911de, line 331: Nit: Formatting of `<1>` isn't aligned with `<2>`

mp911de (changes requested):

PR #3638, opened 2024-10-12
little bit refactoring

jack5505:


mp911de:
What exactly is the pull request supposed to fix? Care to elaborate?

mp911de:
Our API is non-null by default. There's no reason to introduce Jakarta Validation API to our infrastructure components.

PR #3642, opened 2024-10-16
Add config option to disable query validation

### LINKED ISSUE #1690: Query validation significantly slows down application startup  [DATAJPA-1370]
**[Сергей Цыпанов](https://jira.spring.io/secure/ViewProfile.jspa?name=stsypanov)** opened **[DATAJPA-1370](https://jira.spring.io/browse/DATAJPA-1370?redirect=false)** and commented

Having dozens of interfaces extending JpaRepository and several hundreds of methods annotated with `@Query` it takes application significant type to start up.

I think there could be an option to disable validation e.g. for tests.

Also SimpleJpaQuery::validateQuery can be improved: e.g. we should compute error message only if necessary

---

2 votes, 4 watchers

christophstrobl:
Resolves: #1690

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/config/JpaRepositoryConfigExtension.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/config/JpaRepositoryConfigExtension.java
@@ -83,6 +83,7 @@ public class JpaRepositoryConfigExtension extends RepositoryConfigurationExtensi
 	private static final String ENABLE_DEFAULT_TRANSACTIONS_ATTRIBUTE = "enableDefaultTransactions";
 	private static final String JPA_METAMODEL_CACHE_CLEANUP_CLASSNAME = "org.springframework.data.jpa.util.JpaMetamodelCacheCleanup";
 	private static final String ESCAPE_CHARACTER_PROPERTY = "escapeCharacter";
+	private static final String QUERY_VALIDATION = "queryValidation";
 
 	private final Map<Object, String> entityManagerRefs = new LinkedHashMap<>();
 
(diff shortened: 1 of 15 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/config/JpaRepositoryConfigExtension.java
```diff
@@ -118,6 +119,7 @@ public void postProcess(BeanDefinitionBuilder builder, RepositoryConfigurationSo
 		builder.addPropertyValue("transactionManager", transactionManagerRef.orElse(DEFAULT_TRANSACTION_MANAGER_BEAN_NAME));
 		builder.addPropertyReference("entityManager", entityManagerRefs.get(source));
 		builder.addPropertyValue(ESCAPE_CHARACTER_PROPERTY, getEscapeCharacter(source).orElse('\\'));
+		getQueryValidation(source).ifPresent(it -> builder.addPropertyValue(QUERY_VALIDATION, it));
```
mp911de, line 122: We don't need that one since we're not updating `@EnableJpaRepositories`

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/SimpleJpaQuery.java
```diff
@@ -45,8 +46,8 @@ final class SimpleJpaQuery extends AbstractStringBasedJpaQuery {
 	 * @param valueExpressionDelegate must not be {@literal null}
 	 */
 	public SimpleJpaQuery(JpaQueryMethod method, EntityManager em, @Nullable String countQueryString,
-			QueryRewriter queryRewriter, ValueExpressionDelegate valueExpressionDelegate) {
-		this(method, em, method.getRequiredAnnotatedQuery(), countQueryString, queryRewriter, valueExpressionDelegate);
+			QueryRewriter queryRewriter, ValueExpressionDelegate valueExpressionDelegate, boolean validation) {
+		this(method, em, method.getRequiredAnnotatedQuery(), countQueryString, queryRewriter, valueExpressionDelegate, validation);
```
mp911de, line 50: Probably a `QueryValidator` object would make sense instead of a magic boolean. I can handle that during the merge.

PR #3648, opened 2024-10-22
Replace `!Optional.isPresent()` with `isEmpty()` for readability

jeonghyeon00:
Refactor: replace !optionalValue.isPresent() with optionalValue.isEmpty()


mp911de:

Issue #3649, opened 2024-10-25
NPE in HqlQueryTransformer.isSubquery for UPDATE HQL


dforrest-es:
With 3.2.11, repository initialization fails with
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.antlr.v4.runtime.ParserRuleContext.getParent()" because "ctx" is null
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.isSubquery(HqlQueryTransformer.java:105) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.isSubquery(HqlQueryTransformer.java:105) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.isSubquery(HqlQueryTransformer.java:105) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.isSubquery(HqlQueryTransformer.java:105) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.isSubquery(HqlQueryTransformer.java:105) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.isSubquery(HqlQueryTransformer.java:105) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryTransformer.visitVariable(HqlQueryTransformer.java:362) ~[spring-data-jpa-3.2.11.jar:3.2.11]
	at org.springframework.data.jpa.repository.query.HqlQueryRenderer.visitVariable(HqlQueryRenderer.java:1) ~[spring-data-jpa-3.2.11.jar:3.2.11]
```

Issue is very similar to #2977 but for update clauses:

```
update MyEntityStaging AS mes
set mes.col = 'test'
where mes.id = 1
```

Entity:
```
@Entity
@NamedQuery(name = "MyEntity.updateCol", query = """
        update MyEntity AS mes
        set mes.col = 'test'
        where mes.id = 1
""")
class MyEntity {
    @Id
    @GeneratedValue(strategy = GenerationType.SEQUENCE)
    private Long id;

    private String col;
}
```

Repository:
```
[...]

PR #3650, opened 2024-10-25
Fix NPE when parsing modifying HQL

christophstrobl:
Resolves: #3649 

Needs to be forward ported to 3.3.x and _main_.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
@@ -106,7 +106,7 @@ class HqlQueryTransformer extends HqlQueryRenderer {
 		} else if (ctx instanceof HqlParser.UpdateStatementContext) {
 				return false;
 		} else {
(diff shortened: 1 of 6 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlQueryTransformer.java
```diff
@@ -101,6 +101,10 @@ private static boolean isSubquery(ParserRuleContext ctx) {
 			return false;
 		} else if (ctx instanceof HqlParser.InsertStatementContext) {
 			return false;
+		} else if (ctx instanceof HqlParser.DeleteStatementContext) {
+			return false;
+		} else if (ctx instanceof HqlParser.UpdateStatementContext) {
+				return false;
```
mp911de, line 107: We should add another guard in case the parent is empty to eliminate this bug class for future revisions.

christophstrobl: yeah - sounds good - let me do that

PR #3654, opened 2024-10-28
Rewrite String queries for DTO Constructor Expressions

### LINKED ISSUE #2327: Implement class-based DTO projections for Fluent API
#2294 introduced the Fluent API. However, scheduling issues prevented us from implementing class-based DTOs.

As part of that development effort, the following chunk of code from `FetchableFluentQueryByExample` was withdrawn since it isn't relevant to everything.

```java
	/**
	 * Draft version of a projection-based query using class-based DTOs.
	 *
	 * @param sort
	 * @param queryType
	 * @param example
	 * @return
	 */
	private TypedQuery<R> createProjectionQueryByExample(Sort sort, Class<R> queryType, Example<S> example) {

		CriteriaBuilder builder = this.entityManager.getCriteriaBuilder();
		CriteriaQuery<R> query = builder.createQuery(queryType);

		Root<R> root = query.from(queryType);
		query.select(root);

		Predicate predicate = QueryByExamplePredicateBuilder.getPredicate(
				builder.createQuery(example.getProbeType()).from(example.getProbeType()), builder, example, escapeCharacter);

		if (predicate != null) {
			query.where(predicate);
		}

		if (sort.isSorted()) {
			query.orderBy(toOrders(sort, root, builder));
		}

		return this.entityManager.createQuery(query);
	}
```

### LINKED ISSUE #3076: Rewrite string-queries to use constructor expressions when return type is DTO
Consider the following query:

```java
@Entity
class Person { … }

class PersonProjection { … }


@Query("SELECT p FROM Person p")
PersonProjection findBy(…)
```

The query selects items from the Person entity while its return type is a DTO projection.

It would be good to rewrite such queries to use DTO projections for matching properties along the lines of:

```java
@Query("SELECT new com.acme.PersonProjection(p.firstName, p.lastName) FROM Person p")
PersonProjection findBy(…)
```

### LINKED ISSUE #3702: Class-based DTOs are not yet supported.
springboot version: v3.3.7-SNAPSHOT

```
Page<SchoolInterface> pageResult = schoolRepository.findBy(specification, q -> q.as(SchoolInterface.class));
```
When projection target is Interface, it works well.

But error occurred when I change the projection target to DTOs(normal class or record type)

java.lang.UnsupportedOperationException: Class-based DTOs are not yet supported.

I hope to know why must use interface.

mp911de:
Closes #3076
Closes #2327
Closes #3702

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JakartaTuple.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JakartaTuple.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.jpa.repository.support;
+
+import jakarta.persistence.Tuple;
+
+import java.util.Arrays;
+import java.util.List;
+
+import com.querydsl.core.types.Expression;
+import com.querydsl.core.types.ExpressionBase;
+import com.querydsl.core.types.ExpressionUtils;
+import com.querydsl.core.types.FactoryExpression;
+import com.querydsl.core.types.Ops;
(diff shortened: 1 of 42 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JakartaTuple.java
```diff
@@ -0,0 +1,85 @@
...
+		return (R) this;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (obj == this) {
+			return true;
+		} else if (obj instanceof FactoryExpression) {
+			FactoryExpression<?> c = (FactoryExpression<?>) obj;
+			return args.equals(c.getArgs()) && getType().equals(c.getType());
```
Seol-JY, line 79:
```suggestion
                } else if (obj instanceof FactoryExpression<?> c) {
                    return args.equals(c.getArgs()) && getType().equals(c.getType());
```

Issue #3682, opened 2024-11-17
Empty attributePaths on @EntityGraph produces exception


lbruun:
From how I understand the JPA specs there would be nothing wrong with the following Spring Data JPA annotation:

```java
  @EntityGraph(attributePaths = {}) // Purpose: exclude all EAGER relations
```

The purpose would be to create a query using an entity graph with no attributes added. Perfectly legal in JPA. (and indeed a valid use case)

However, with Spring Data JPA this is not possible as class `JpaEntityGraph` uses the following (flawed?) logic to distinguish between dynamic and static named graphs:

```java
public boolean isAdHocEntityGraph() {
    return !attributePaths.isEmpty();
}
```

Therefore, the query result in an exception :

```text
java.lang.IllegalArgumentException: The given JpaEntityGraph [...] is not dynamic
```

This is wrong. The entity graph is indeed dynamic. It just happens to be an entity graph with no attributes added. 

Another way to phrase the problem is that putting this on a query method in a Spring Data repo:

```java
  @EntityGraph(attributePaths = {}) 
```

should NOT result in an exception, but should result in a query being performed using an JPA entity graph with no attributes added. Alternatively, please provide another way to achieve the goal.

Thanks.

christophstrobl:
Thank you for the heads up @lbruun. True, you could have a entity graph without any attributes. 
Let me see what we can do to lift the existing restriction.

PR #3684, opened 2024-11-20
Lift restriction on empty `JpaEntityGraph` attributes

christophstrobl:
Closes: #3682


mp911de:

Issue #3672, opened 2024-11-13
Upgrade to Eclipselink 5.0.0-B05


mp911de:

mp911de:
Failing tests:

```
Syntax error parsing [select concat(?1,u.id,?2) as id from User u]. 
[31, 33] The identification variable 'id' cannot be a reserved word. (select concat(?1, u.id, ?2) as  [ id ] ...
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.createQuery(EntityManagerImpl.java:1848)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.springframework.orm.jpa.ExtendedEntityManagerCreator$ExtendedEntityManagerInvocationHandler.invoke(ExtendedEntityManagerCreator.java:364)
	at jdk.proxy2/jdk.proxy2.$Proxy202.createQuery(Unknown Source)
	at org.springframework.data.jpa.repository.query.SimpleJpaQuery.validateQuery(SimpleJpaQuery.java:91)
	... 53 more
```

Issue #3587, opened 2024-08-21
Consider `NULLS` precedence using `Sort` for Criteria Queries


mp911de:
See also #1280

Issue #3136, opened 2023-09-01
Update JPQL and HQL parsers with newly introduced JPA 3.2 keywords


mp911de:
See https://deploy-preview-652--jakartaee-specifications.netlify.app/specifications/persistence/3.2/

* Adds `union`, `intersect`, `except`, `cast`, `left`, `right`, and `replace` for Jakarta Persistence QL and criteria queries
* Adds `||` string concatenation operator to Jakarta Persistence QL
* Adds support for specifying null precedence when ordering Jakarta Persistence QL and criteria queries

Nendanfito:
Does null precedence work on Page requests? Because its not working for me with version 3.2.1

`
Sort.Order sort = switch (direction != null ? direction : "") {
            case "asc" -> Sort.Order.asc(sortBy);
            case "desc" -> Sort.Order.desc(sortBy).nullsLast();
            default -> Sort.Order.asc("code");
        };

PageRequest pageRequest = PageRequest.of(
                alarmPageableFilter.getPageNumber(),
                alarmPageableFilter.getPageSize(),
                Sort.by(sort));
`

mp911de:
No, it's not working with 3.2.1 because the JPA spec level is at 3.1. Starting with JPA 3.2, the API will specify null precedence (see https://projects.eclipse.org/projects/ee4j.jpa/releases/3.2)

ducanh2110:
I have the same issue too. Dont understand why it's happen

joseardaiz:
@mp911de and @christophstrobl - would you guys know when the issue that @Nendanfito explained here will be fixed?

Nendanfito:
@emaranga I configured it  for spring in general in the application.yaml:
```
spring:
  jpa:
    properties:
      hibernate:
        order_by:
          default_null_ordering: last
```
With that solution all the querys made by jpa will return nulls last, in case you don't want it you have to made a nativeQuery.

PR #3695, opened 2024-11-29
Upgrade to JPA 3.2

mp911de:
- [x] #3671
- [x] #3672
- [x] #3587
- [x] #3136
- [x] #3701

### DIFF
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java
@@ -544,7 +544,7 @@ public interface UserRepository extends JpaRepository<User, Integer>, JpaSpecifi
 
 	List<RolesAndFirstname> findRolesAndFirstnameBy();
 
-	@Query(value = "FROM User u")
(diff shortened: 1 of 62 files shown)


spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java
```diff
@@ -544,7 +544,7 @@ List<User> findUsersByFirstnameForSpELExpressionWithParameterIndexOnlyWithEntity
 
 	List<RolesAndFirstname> findRolesAndFirstnameBy();
 
-	@Query(value = "FROM User u")
+	@Query(value = "SELECT u FROM User u")
```
christophstrobl, line 547: I think we should not alter the query here since it's valid JPQL even without the `SELECT` just because the Eclipselink preview is unable to deal with it. 
```
Problem compiling [FROM User u]
[7, 11] The identification variable 'this' is not defined in the FROM clause. (SELECT  [ this ] ...
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildException(HermesParser.java:175)
```

PR #3698, opened 2024-12-02
Extract repeated string literals in `SimpleJpaRepository` to constants

dukbong:
This commit replaces repeated string literals with constants, which is expected to improve the maintainability of the code.

In addition to this commit, I was considering committing a change to overload the getQueryString method and apply lazy evaluation. However, I'm unsure if this change would actually provide any significant benefits. I would like to know your thoughts on this approach.


mp911de:

PR #3715, opened 2024-12-13
refactor: Improve exists check in JpaQueryExecution

minwoo1999:
### Changes
Updated getResultList().isEmpty() to getResultStream().findFirst().isPresent() for checking query existence.
Improved performance by processing only the first result.

### Reason

It seems that this method is only meant to check if the query exists. Therefore, fetching the entire query result does not seem necessary.
Avoids loading all results into memory, reducing overhead and improving efficiency.

### Testing
Verified with existing test cases, ensuring identical results.

### Review Request
Please review and share any feedback. 🙇‍♂️


mp911de:
`ExistsExecution` is only used for derived queries where we set `query.setMaxResults(1)`. Therefore, instantiating a Java 8 stream instead of reading one result into a `List` is almost the same overhead. We want to keep our code as-is.

PR #3727, opened 2024-12-28
Support custom countSpec in SimpleJpaRepository.findAll

JoshuaChen:
In some performance optimization scenarios, a custom count spec is needed to improve query performance.


mp911de:
Thanks for your suggestion. We're going to include it. However, next time, please squash your commits into one and make sure to include tests.

mp911de:

PR #3730, opened 2025-01-07
Use `*` instead of primary alias in count queries with CTE

christophstrobl:
Still a draft - might be worth exploring if we can maybe avoid the mutable state in the count transformer that sets the CTE flag.


mp911de:
That's merged, backported, and polished now.

PR #3766, opened 2025-02-09
Refactor string constants.

JunggiKim:
# Description
Refactored string literals to constants in NamedQuery class to improve code maintainability and reduce duplication.

# Changes
- Extracted commonly used string literals ("NativeQuery", "Query") to static final constants
- Updated all occurrences to use the new constants
- No functional changes or behavior modifications


mp911de:
Thanks for reaching out. We do want to keep that code as-is. Introducing indirections for a few constants causes more mental overhead than it attempts to fix.

PR #3783, opened 2025-02-20
 Change the if-else structure to early return method

mini0192:
I checked the code and confirmed that if-else was not an early return method, so I fixed it.


mini0192:
> There's nothing really to fix as it isn't broken. Also, the PR contains much more commits than necessary, especially the POM changes.

Does this mean it is no need to merge?

mp911de:
Exactly.

PR #3794, opened 2025-02-28
Fix JoinType on associations when using MatchMode.ANY

### LINKED ISSUE #3763: Invalid JoinType in QueryByExamplePredicateBuilder when using MatchMode ANY
Hi,

I stumbled upon a bug when using `Example` to find objects when matching with both a primitive attribute and an association at the same time when using `MatchMode.ANY`.

The join should be a left join when using `MatchMode.ANY` and only an inner join when using `MatchMode.ALL`.

Here is a reproducer :
[https://github.com/ArnaudLec/query-by-example-join-match-any-bug](https://github.com/ArnaudLec/query-by-example-join-match-any-bug)

ArnaudLec:
Fixes #3763 


christophstrobl:
Thank you @ArnaudLec! Merged to _main_ and _3.4.x_ development lines.

PR #3865, opened 2025-04-30
Fix entity-persistence.adoc.

w3n9:


schauder:
Thanks for improving that example.

This is merged.

For any observers: the original example was still fine, since the isNew check happens before the lifecycle events, but using `postpersist` is certainly less confusing.

PR #3868, opened 2025-05-05
Explore Search Results

mp911de:
Return `SearchResult` from Repository Query methods.

Depends on spring-projects/spring-data-commons#3285


christophstrobl:
Merged to _4.0.x_.

Issue #3425, opened 2024-04-12
Using native Hibernate multitenancy with Spring Data JPA breaks on application boot


mnnayeck:
I have configured Hibernate's native multitenancy in the attached persistence-tenant.txt.
As you can see: 

`			<property name="hibernate.multiTenancy" value="DATABASE" />
            <property name="hibernate.multi_tenant_connection_provider" value="mu.maccs.ffa.dao.spi.MultiTenantConnectionProviderImpl" />
            <property name="hibernate.tenant_identifier_resolver" value="mu.maccs.ffa.dao.spi.MultiTenantIdentifierResolverImpl" />
`

I am using a simple JPARepository class as attached.

The problem is that on bootstrapping, org.springframework.data.jpa.repository.support.JpaRepositoryFactory's constructor makes the following call:
`this.extractor = PersistenceProvider.fromEntityManager(entityManager);`

which actually creates a hibernate sessionImpl to determine the PersistenceProvider to use in the JPARepository.

This is **wrong** because since my application is not executing any database call on boot, a Hibernate session creation is FORCED (which in turns creates a connection with the underlying datasource for that particular tenant).  This line is the culprit:
`Class<?> entityManagerType = em.getDelegate().getClass();`

Because of this, I am forced to set a default tenantIdentifier in MultiTenantIdentifierResolverImpl which isn't right. There should be NO tenant resolution on creation of JpaRepositoryFactory.
[persistence-tenant.txt](https://github.com/spring-projects/spring-data-jpa/files/14963238/persistence-tenant.txt)
[UserRepository.txt](https://github.com/spring-projects/spring-data-jpa/files/14963256/UserRepository.txt)
[MultiTenantConnectionProviderImpl.txt](https://github.com/spring-projects/spring-data-jpa/files/14963264/MultiTenantConnectionProviderImpl.txt)
[MultiTenantIdentifierResolverImpl.txt](https://github.com/spring-projects/spring-data-jpa/files/14963265/MultiTenantIdentifierResolverImpl.txt)

Log Trace:
`
2024-04-13 00:19:20,870 [main] TRACE org.springframework.beans.factory.support.DefaultListableBeanFactory.invokeInitMethods:1831 - Invoking afterPropertiesSet() on bean with name 'userRepository'
[...]

PR #3885, opened 2025-05-15
Avoid early `EntityManager` initialization during `PersistenceProvider` lookup

ari-spf:
Changes into the mechanism of `PersistenceProvider.fromEntityManager` to prevent the initialization of a hibernate session which would result in an exception on boot if multi-tenancy is enabled and no tenant context information has been provided.

This is a fix for bug: https://github.com/spring-projects/spring-data-jpa/issues/3425
Existing test cases were adjusted to account for the new resolution strategy.

PS: If this change gets accepted, would it be possible to cherry-pick the fix to a patch version?

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java
@@ -56,6 +56,7 @@ import org.springframework.util.ConcurrentReferenceHashMap;
  * @author Jens Schauder
  * @author Greg Turnquist
  * @author Yuriy Tsarkov
+ * @author Ariel Morelli Andres (Atlassian US, Inc.)
  */
 public enum PersistenceProvider implements QueryExtractor, ProxyIdAccessor, QueryComment {
 
@@ -316,7 +317,7 @@ public enum PersistenceProvider implements QueryExtractor, ProxyIdAccessor, Quer
 	}
 
 	/**
-	 * Determines the {@link PersistenceProvider} from the given {@link EntityManager}. If no special one can be
+	 * Determines the {@link PersistenceProvider} from the given {@link EntityManagerFactory}. If no special one can be
 	 * determined {@link #GENERIC_JPA} will be returned.
 	 *
 	 * @param emf must not be {@literal null}.
@@ -324,7 +325,7 @@ public enum PersistenceProvider implements QueryExtractor, ProxyIdAccessor, Quer
 	 */
 	public static PersistenceProvider fromEntityManagerFactory(EntityManagerFactory emf) {
 
-		Assert.notNull(emf, "EntityManager must not be null");
+		Assert.notNull(emf, "EntityManagerFactory must not be null");
 
 		Class<?> entityManagerType = emf.getPersistenceUnitUtil().getClass();
 		PersistenceProvider cachedProvider = CACHE.get(entityManagerType);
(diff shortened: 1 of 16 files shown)


mp911de:
I'm not sure this change is complete because there are several classes calling `PersistenceProvider.fromEntityManager`. Also, this change is a breaking one as `PersistenceProvider` is a public class. I'd appreciate if you have a look at the `4.0.x` change of [`PersistenceProvider`](https://github.com/spring-projects/spring-data-jpa/blob/4.0.x/spring-data-jpa/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java#L325). Feel free to update your pull request to introduce the static `fromEntityManagerFactory(…)` method and update all calling code to the new method to verify that the application startup isn't affected using the `EntityManagerFactory` approach.

ari-spf:
Thanks @mp911de for reviewing the code. I'd. like to understand this comment more:
> I'm not sure this change is complete because there are several classes calling PersistenceProvider.fromEntityManager. Also, this change is a breaking one as PersistenceProvider is a public class.

I've paid special attention to not change the signature of the methods of the PersistenceProvider, only it's logic to resolve it from the `EntityManager`. All callers should be able to utilize the new behavior. Would you mind elaborating a bit more on where it would break in other usages?


With regards of the `4.0.x` code that you've pointed out, is there any reason why the method `fromEntityManager(EntityManager em)` isn't just returning `fromEntityManagerFactory(em.getEntityManagerFactory())` ?
Moreover, I'm not sure [this 4.0.x code](https://github.com/spring-projects/spring-data-jpa/blob/4.0.x/spring-data-jpa/src/main/java/org/springframework/data/jpa/provider/PersistenceProvider.java#L325C36-L346)  would be able to correctly infer the `PersistenceProvider` when a proxy created by the [`AbstractEntityManagerFactoryBean`](https://github.com/spring-projects/spring-framework/blob/main/spring-orm/src/main/java/org/springframework/orm/jpa/AbstractEntityManagerFactoryBean.java#L440-L470) is passed.

mp911de:
Interacting with the EntityManager always uses the proxy. getEntityManagerFactory() doesn’t create the EntityManager, so we can use this optimization.

Sometimes, we already have the EntityManagerFactory in place (or the metamodel) for 4.0 functionality.

You’re right about PersistenceProvider. I have misread the changes with JpaClassUtils.

The 4.0 code is missing the proxy handling that your PR has introduced.

Let me have another look tomorrow.
Am 15. Mai 2025, 20:05 +0200 schrieb ari ***@***.***>:

ari-spf:
thanks so much @mp911de for replying on your evening! Let's chat tomorrow.

ari-spf:
@mp911de I've re-implemented this taking into account the changes that happened for the branch 4.0.x which was merged to main.
Can you please take another look?

ari-spf:
@mp911de I'm making one last attempt to bring this back to the top of your inbox.

mp911de:
Just because we don't reply timely doesn't mean that things aren't in our field of sight. We're currently working towards fixing regressions to be released by next week. With a limited number of folks who can work on tickets, features naturally don't have the highest priority in that phase.

ari-spf:
Thanks for your reply, that helps me understand the timing. Kindly let me know if there is something I can help with. I'll be out for 2 weeks, but happy to help once I'm back.

mp911de:
Thank you for your contribution. That's merged, polished, and backported now. Using your previous approach made more sense and I've removed checks to the `EntityManager` interface (and `PersistenceUnitUtil`) entirely. I rolled back unrelated changes in your commit and updated the author style.

Issue #2761, opened 2023-01-12
Specification<T>.toPredicate(...) Change type of first parameter to From<?, T>


jobayle:
Dear spring devs,

I'd like to suggest the following change in interface `Specification<T>`:

`Predicate toPredicate(Root<T> root, CriteriaQuery<?> query, CriteriaBuilder criteriaBuilder);`

to

`Predicate toPredicate(From<?, T> root, CriteriaQuery<?> query, CriteriaBuilder criteriaBuilder);`

Because `From<T, U>` is the supertype of both `Root<T>` and `Join<T, U>`, it would allow better reuse of specifications when joining tables.

----

Our use case is the following, we have a OneToMany relation
Table CarMaker <>---> Table Car

We have already made Specification factory methods from Request DTOs for entities `CarMaker` and `Car`.
We would like to reuse these 2 methods when filtering on both CarMaker and Car:

```java
Specification<CarMaker> filterCarMakers(CarMakerRequestDTO request) { ... }

Specification<Car> filterCars(CarRequestDTO request) { ... }

Specification<CarMaker> filterCarMakersAndCars(CarMakerAndCarAggregatedRequestDTO  request) {
    return filterCarMakers(request.getCarMakerRequest())
            .and((r, q, c) -> filterCars(request.getCarRequest()).toPredicate(r.join("cars", JoinType.INNER), q, c));
}
```
see also: https://stackoverflow.com/a/46955740/6705221

mp911de:
From an API perspective, `Root` declares one method more than `From` does, and it denotes a more specific type. Generally speaking, `From` could be a better alternative than `Root` to improve reuse of specifications. However, specifications weren't designed with reuse in mind and we have no way to introduce a non-breaking change considering the amount of existing specification implementations.

This could be a topic worth revisiting for Spring Data JPA 4.0.

mp911de:
Another approach here could be the introduction of a functional-style composition API to build and reuse specification-like fragments. Anyone willing to experiment with that kind of approach can take a spike and we're happy to keep the discussion going.

bukajsytlos:
when we are at it, why not `Path<X>`?

svenmeier:
I've put together a possible implementation based on how we handle it in our project. Due to limitations with modifying Spring classes, our local solution includes a bit more code duplication.

[...]

PR #3919, opened 2025-06-13
Issue/2761 specifications with from

svenmeier:
Proposal for GH-2761

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/JoinSpecification.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/JoinSpecification.java
@@ -1,0 +1,35 @@
+package org.springframework.data.jpa.domain;
+
+import jakarta.persistence.criteria.*;
+
+/**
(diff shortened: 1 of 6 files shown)


mp911de:
For 4.0, we've introduced `PredicateSpecification` for improved composition, therefore, we don't want to include another nestable specification. The Join-specification seems rather specific and given the improved flexibility with `From`, we do not want to introduce another specification variant.

Farley-Chen:
> For 4.0, we've introduced `PredicateSpecification` for improved composition, therefore, we don't want to include another nestable specification. The Join-specification seems rather specific and given the improved flexibility with `From`, we do not want to introduce another specification variant.

Hi, I have found that in `PredicateSpecification` there is still using `Root`, is there any plan or solution to use `From` for `join` purpose ?

mp911de:
Good point @Farley-Chen. We've revised `PredicateSpecification` by apply `<T>` to `From<?, T>` for improved composition. While we lose a bit of type information, we also drop our opinion on type restrictions and open up the model for improved composition and reuse across e.g. Joins.

PR #3920, opened 2025-06-15
Cache query strings in SimpleJpaRepository to improve performance

Dockerel:
Optimizes SimpleJpaRepository by caching the query strings used for deleteAll and count operations.

- Introduced final fields to cache the deleteAll and count query strings during repository construction.
- Updated usages in deleteAllInBatch() and count() to use the cached query strings instead of generating them on every call.

Previously, the query strings for deleteAll and count were generated on each method invocation, causing repeated ` operations and unnecessary object creation. By caching these strings, we reduce redundant computation and improve performance, especially in applications with frequent batch delete or count operations.

The change would be safe as the query strings are based on immutable entity metadata and do not change during the repository's lifecycle.

Please review if there are scenarios where the entity metadata might change after repository construction, which could affect the correctness of the cached queries.

Let me know if you have any concerns or suggestions for further improvement!


mp911de:

PR #3932, opened 2025-07-09
Replace regex with startsWith / endsWith check for LIKE pattern detection

Dockerel:
Replace regex `%.*%` with `startsWith` / `endsWith` check for LIKE pattern detection
- Removes regex overhead and Micro-performance gain
- Explicit intent (startsWith/endsWith) is clearer than a regex.

Previously, each call to expression.matches("%.*%") incurred regex compilation and Matcher allocation overhead. Caching the pattern with

```java
private static final Pattern CONTAINING = Pattern.compile("%.*%");
```

would avoid recompilation, but switching to startsWith("%") && endsWith("%") is even more efficient and more readable, so the implementation has been changed accordingly.

Let me know if you have any concerns or suggestions for further improvement.
Thank you for your time and consideration!


mp911de:

PR #3967, opened 2025-08-08
Removed throw Exception code that does not occur

Tussle0410:
Description
---
When selecting `QueryLookupStrategy` using a `switch` statement with Enum-defined values, if null is passed, it defaults to `CreateIfNotFoundQueryLookupStrategy`, which means the switch statement's default case cannot occur. Therefore, I determined that the corresponding throw Exception would never be executed.

So, I removed the code that performs the throw Exception.

Changes
---
Removed the throw exception part in the `default` case of the `switch` statement.


mp911de:
The enum comes from a different library and therefore, changes to the enum aren't directly coupled to the Spring Data JPA module. The exception is more of a safe-guard to fail when something unexpected will be passed in. Therefore, we want to keep the exception arrangement as-is.

Tussle0410:
It seems I hadn't considered that it could serve as a safeguard against unexpected errors.
Thank you for providing such a good response with your opinion.

PR #3993, opened 2025-08-26
Qualify identifier used in `SimpleJpaRepository.deleteAllByIdInBatch(…)` JPQL

### LINKED ISSUE #3990: EclipseLink fails resolving unqualified identifier property using `IN` on batch delete by Id
Dear spring-data-jpa,

In the following issue, there was 2 issues described (I should have created 2 separate issues). The first issue has been resolved. The second issue is still pending and I would like to have your assessment on it.

https://github.com/spring-projects/spring-data-jpa/issues/3983#issuecomment-3219240960) 

The second issue is with regards to the **deleteAllByIdInBatch()** method in the above example which is causing the following dump:

```
Caused by: Exception [EclipseLink-0] (Eclipse Persistence Services - 4.0.7.v202506240759-923a428d4722d199eff24e730890e156ce7ea9de): org.eclipse.persistence.exceptions.JPQLException
Exception Description: Problem compiling [delete from IssueTable x where id in :ids]. 
[31, 33] The identification variable 'id' is not defined in the FROM clause.
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildException(HermesParser.java:159)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.validate(HermesParser.java:337)
```

The query is created by the Spring Data jpa method  SimpleJpaRepository<T,ID>.deleteAllByIdInBatch(Iterable) which is then calling the  "Query query = entityManager.createQuery(queryString);" with a wrong SQL.

Wrong: "delete from IssueTable x where id in :ids"
Valid: "delete from IssueTable x where x.id in :ids

Here is a simple unit test **testDeleteAllByIdInBatch()** to reproduce the issue (java files + pom) to reproduce the issue.

[pom.xml.txt](https://github.com/user-attachments/files/21972459/pom.xml.txt)
[SpringJPAIssueTest.java](https://github.com/user-attachments/files/21972462/SpringJPAIssueTest.java)
[IssueTable.java](https://github.com/user-attachments/files/21972463/IssueTable.java)
[IssueRepository.java](https://github.com/user-attachments/files/21972461/IssueRepository.java)

Best regards,
Joel

pmh5574:
This PR addresses GH-3990 by adding a regression test and refining the JPQL used for deleteAllByIdInBatch.

### Background
EclipseLink can fail when the JPQL DELETE uses an unqualified state field in the WHERE clause. Qualifying the field with the range variable alias makes the query portable.

### Changes
- Qualify the delete-by-id JPQL:
  `delete from %s x where x.%s in :ids`
- Add `EclipseLinkDeleteAllByIdInBatchSimpleIdTests`:
  - Save 3 users and flush to allocate IDs
  - Delete two using `deleteAllByIdInBatch`
  - Assert the remaining entity is the expected one

### Verification
Built and tested locally with `./mvnw clean install` (JDK 17+).

Closes #3990

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -94,7 +94,7 @@ public abstract class QueryUtils {
 
 	public static final String COUNT_QUERY_STRING = "select count(%s) from %s x";
 	public static final String DELETE_ALL_QUERY_STRING = "delete from %s x";
-	public static final String DELETE_ALL_QUERY_BY_ID_STRING = "delete from %s x where %s in :ids";
+	public static final String DELETE_ALL_QUERY_BY_ID_STRING = "delete from %s x where x.%s in :ids";
 
 	// Used Regex/Unicode categories (see https://www.unicode.org/reports/tr18/#General_Category_Property):
 	// Z Separator
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/support/EclipseLinkDeleteAllByIdInBatchSimpleIdTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/support/EclipseLinkDeleteAllByIdInBatchSimpleIdTests.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2008-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


mp911de:
Thank you for your contribution. That's merged, polished, and backported now. Please review properly AI-generated contributions. In this case, we have already a better place for the tests in `EclipseLinkJpaRepositoryTests` that also indicate that there has been some effort to handle batch deletes. Also, the commit message is just noise and doesn't explain what the commit is supposed to achieve, merely a duplication by summarizing changes textually. Therefore, we've rewritten the commit message.

PR #3998, opened 2025-08-29
Reintroduce Specification.where(Specification)

### LINKED ISSUE #3992: Reintroduce `Specification.where(Specification)`
See #3893 and user feedback:

> Specifically, the static where(...) method now expects a PredicateSpecification<T> instead of a Specification<T>. This causes compile-time errors in existing projects that rely on Specification builders.
> 
> Would it be possible to provide an overload of where(...) that still accepts Specification<T> (delegating internally to PredicateSpecification), or any kind of polymorphic support? This would greatly simplify migration by avoiding a large-scale refactor of existing specifications.

We should consider reintroducing `Specification.where(Specification)` to simplify the migration to Spring Data JPA 4.0.

nam0oo:
Reintroduce the `Specification.where(Specification)` overload to improve the migration path for users upgrading to Spring Data JPA 4.0 and to restore the intuitive fluent API.

Closes #3992 

---

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/Specification.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/Specification.java
@@ -48,6 +48,7 @@ import org.springframework.util.Assert;
  * @author Jens Schauder
  * @author Daniel Shuy
  * @author Sergey Rukin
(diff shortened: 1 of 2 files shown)


mp911de:

PR #4000, opened 2025-09-01
Fix unpaged revision query

### LINKED ISSUE #3999: Revision repository `findRevisions` throws exception when using `Pageable.unpaged()`
When using `Pageable.unpaged()` for `findRevisions` in `RevisionRepository`, this line (and lines after this one) will throw an exception.

https://github.com/spring-projects/spring-data-jpa/blob/f50d3567fe48139282a3eb2b27f0b2b1a1683f55/spring-data-envers/src/main/java/org/springframework/data/envers/repository/support/EnversRevisionRepositoryImpl.java#L185

It would be nice to check if the pageable object is `unpaged` and handle it.

mp911de:
Closes #3999

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa-parent</artifactId>
-	<version>4.0.0-SNAPSHOT</version>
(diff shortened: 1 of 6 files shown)


schauder:
Merged.

PR #4003, opened 2025-09-03
Allow customization of `@RevisionTimestamp` property name

### LINKED ISSUE #2850: Hardcoded revision property?
Why did you hardcode `"timestamp"` as revision property in `org.springframework.data.envers.repository.support.EnversRevisionRepositoryImpl#findLastChangeRevision` ?

My custom revision entity has different `@RevisionTimestamp` property than `timestamp`. 

This results in `org.hibernate.QueryException: could not resolve property: timestamp of:...` when `org.springframework.data.repository.history.RevisionRepository#findLastChangeRevision` is called.

Chaedie:
Related Issue: #2850 

Fixes #2850 by replacing hardcoded "timestamp" property name with dynamic @RevisionTimestamp field detection.


mp911de:
Seems you never ran the tests for the change as the build has failed initially.

PR #4024, opened 2025-09-25
Consistent handling of `Specification.unrestricted()` in `Specification.not(..)`

### LINKED ISSUE #4023: Inconsistency in `Specification.unrestricted()` semantics
### Summary

The semantics of `Specification.unrestricted()` are inconsistent across different logical operator methods (`or`, `and`, `not`, etc.), leading to confusion and potentially incorrect assumptions when composing specifications.

### Background

The documentation describes `Specification.unrestricted()` as:
> Simple static factory method to create a specification matching all objects.

This suggests a mental model akin to:
```java
return (root, query, builder) -> builder.conjunction(); // i.e., always true
```

However, the actual implementation is:
```java
return (root, query, builder) -> null;
```
And per the specification contract:
> Specifications returning `null` are considered to not contribute to the overall predicate and their result is not considered in the final predicate.

This leads to two competing interpretations:
So we have two slightly conflicting definitions:
- Interpretation A: `unrestricted()` behaves like `true`
- Interpretation B: `unrestricted()` behaves like SQL `NULL` or the `Unknown` state in three-valued logic.

### Observed Behaviour
In practice, `unrestricted()` behaves like `NULL` in most logical operators:
```java
Predicate predicate = ... ;
Specification other = (r, q, cb) -> predicate;

assertThat(unrestricted().or(other).toPredicate(root, query, builder))
        .isSameAs(predicate);
assertThat(unrestricted().and(other).toPredicate(root, query, builder))
        .isSameAs(predicate);
assertThat(allOf(unrestricted(), other).toPredicate(root, query, builder))
        .isSameAs(predicate);
assertThat(anyOf(unrestricted(), other).toPredicate(root, query, builder))
        .isSameAs(predicate);
```
This implies that `unrestricted()` is ignored in composition, consistent with the `null` return value.

However, `Specification.not(..)` behaves differently:
```java
assertThat(not(unrestricted()).toPredicate(root, query, builder)).isNull(); // fails
```
Instead of returning `null`, it returns `builder.disjunction()` (i.e., `false`), based on reasoning from #3849 (emphasis mine):
> In the current semantics, if specification.toPredicate() returns null, it means no condition—_which logically translates to an Expression true_.
>
> Thus, in the not() method, we can return criteriaBuilder.disjunction() to represent false, because:
> not(true) → false
This implies that `unrestricted()` is treated as `true` in the context of `not()`, breaking consistency.

### Consequences

This inconsistency breaks logical equivalences. For example:
```java
Specification a = Specification.unrestricted();
Specification b = ...;

Specification c = not(a.or(b));
Specification d = not(a).and(not(b));
```
Logically, `c` and `d` should be equivalent (`!(a || b) == !a && !b`), but:

- `c` becomes `not(b)`
- `d` becomes `false`

### Option 1: Document the inconsistency
Clarify in the documentation that `Specification.unrestricted()` behaves like `NULL` in most logical operations, but is treated as `true` in `not()`. This avoids breaking changes but requires users to be cautious when composing specifications.

### Option 2: Make `not(unrestricted())` return `unrestricted()`
This would make `unrestricted()` consistently behave like `NULL`, preserving logical equivalences. However, it breaks existing behaviour where `not(unrestricted())` is expected to match no results.

### Option 3: Make `or()` and `anyOf()` Treat `unrestricted()` as `true`
This aligns with the idea that `unrestricted()` matches all objects. It would make `unrestricted().or(other)` behave like `true || other`, but introduces broader breaking changes.

### Request for Feedback
Would the maintainers prefer Option 1 (document), Option 2 (change `not()`), Option 3 (change `or()`/`anyOf()`), or another approach?
Happy to submit a PR based on the preferred direction.

peteraisher:
### Summary
This PR updates the handling of `Specification.unrestricted()` (and other specifications whose `toPredicate(..)` method returns `null`) in `Specification.not(..)` to ensure consistent handling across all logical operations:

```
unrestricted().or(other)       // equivalent to other
not(unrestricted()).or(other)  // equivalent to other               *changed*
unrestricted().and(other)      // equivalent to other
not(unrestricted()).and(other) // equivalent to other               *changed*
not(unrestricted())            // equivalent to unrestricted()      *changed*
```

### Changes
- Updated tests to verify that `not(unrestricted()).toPredicate(..)` returns `null` and that no `CriteriaBuilder` methods are called.
- Updated implementation accordingly.
- Updated documentation to better explain the semantics of a specification whose `toPredicate(..)` method returns `null`, in particular the static `unrestricted()` factory method.
- Mirrored changes across `Specification`, `DeleteSpecification`, `PredicateSpecification`, and `UpdateSpecification` classes.

### PR Checklist
- [x] I have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] I used the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to my changes.
- [x] I am submitting updated test cases that back my changes.
- [x] I added myself as author in the headers of the classes I touched, and amended the date range in the Apache license headers where needed.

Closes #4023

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/DeleteSpecification.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/domain/DeleteSpecification.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2024-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


mp911de:

PR #4026, opened 2025-09-25
Refactor error message formatting in BadJpqlGrammarErrorListener

jbh010204:
This pull request refactors the error message formatting logic in `BadJpqlGrammarErrorListener` to improve readability, maintainability, and robustness. The main changes involve switching from string concatenation to using a `StringBuilder`, introducing a helper method for line extraction, and refining how error messages are constructed for different types of exceptions.

**Error message formatting improvements:**

* Refactored the `formatMessage` method to use a `StringBuilder` instead of string concatenation, making the code more efficient and easier to read.
* Improved handling of error messages for `NoViableAltException` by extracting the relevant line from the query and inserting a marker at the offending character position, using the new `getLineAt` helper method.
* Added the private static `getLineAt` method to safely retrieve the appropriate line from the query string, handling edge cases where the line number may be out of bounds.
* Enhanced error message construction for `InputMismatchException` by clarifying the expected tokens in the output.

- [ ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/BadJpqlGrammarErrorListener.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/BadJpqlGrammarErrorListener.java
@@ -57,41 +57,75 @@ class BadJpqlGrammarErrorListener extends BaseErrorListener {
 	/**
 	 * Rewrite the error message.
 	 */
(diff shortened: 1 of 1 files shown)


mp911de:
> to improve readability, maintainability

The new variant is less readable and it doesn't fix any problem therefore we want to keep the code as-is.

Issue #4034, opened 2025-10-06
Text block queries are stored twice in hibernate HQL interpretation cache


mdaul-tgtg:
Given:

A text block query in a JPA repository, which does not use pagination:

```
@Repository
class MyRepository extends JpaRepository<MyObject, Long> {

  @Query("""
      SELECT SUM(o.amount)
      FROM MyObject o
      WHERE o.userId = :userId""")
  Optional<Long> sumAmountByUserId(Long userId);
}
```

Then:

On application startup, when the repository bean is initialized, the query is validated, which includes HQL interpretation. It is stored in hibernate's `QueryInterpretationCacheStandardImpl.hqlInterpretationCache` with the raw query as cache key, including spaces and line breaks: `"SELECT SUM(o.amount)\nFROM MyObject o\nWHERE o.userId = :userId"`.

On the first execution of the query, it is rewritten to include the pagination and the resulting HQL is interpreted. The rewritten query is identical to the original since there is no additional order clause from pagination, however, `UnsortedCachingQuerySortRewriter` removes extra spaces and line breaks. That means that `QueryInterpretationCacheStandardImpl` does not recognize the existing query in the interpretation cache, it parses HQL again and stores it with the new cache key `"SELECT SUM(o.amount) FROM MyObject o WHERE o.userId = :userId"`.

While this is not preventing the query from being executed properly, this is inefficient.

Possible solutions:

- Sanitize the query before validation on startup. For example, `JpaQueryLookupStrategy.resolveQuery` could sanitize the query string that is read from the annotation.
- Apply the `QueryRewriter` to the validated query in the `SimpleJpaQuery` constructor. Then we could write our own query rewriter that sanitizes the block, while the default behaviour with `IdentityQueryRewriter` would not change.

bandalgomsu:
I'm interested in this issue and would like to contribute a PR to address it, if no one is already working on it.

mp911de:
Feel free to have a look, mostly analyzing what part of query creation/validation causes this issue.
Am 9. Okt. 2025, 12:57 +0200 schrieb 고수 ***@***.***>:

bandalgomsu:
PR :#4038 

Thanks 🙂

mp911de:
That's something we would fix in the 4.0 line only as the architecture in 3.5.x and earlier doesn't allow a fix without breaking potentially existing functionality.

PR #4038, opened 2025-10-12
Use rewritten EntityQuery for query validation to ensure normalization

bandalgomsu:
Since queryEnhance.rewrite() sanitizes the queryString,
the validation phase has been improved to use the rewritten queryString (= queryEnhance.rewrite()) when creating the validatingEm query.

https://github.com/spring-projects/spring-data-jpa/issues/4034


mp911de:
We've gone for a simpler approach that aligns with the API design and compute the normalized query from `EntityQuery`. Thanks for your suggestion in any case.

Issue #4047, opened 2025-10-17
Refactor QueryTransformers to use constant-first equals for improved null safety


namest504:
In the QueryTransformers.CountSelectionTokenStream class, several .equals() calls are made with a variable on the left and a constant on the right (e.g., token.equals(CONSTANT)).

Example:
```java

class QueryTokens {
    // ...
    static final QueryToken TOKEN_COMMA = token(", ");
    static final QueryToken TOKEN_AS = expression("AS");
    // ...
}
```

QueryTransformers.java
Before:
```java
if (token.equals(TOKEN_AS)) {
    // ...
}
```

After:
```java
if (token.equals(TOKEN_AS)) {
     // ...
}
```

While there is no evidence that token can currently be null, it is a widely accepted Java best practice to place
the constant or literal on the left side of an equals() comparison (e.g., CONSTANT.equals(token)).

This defensively prevents any potential NullPointerException if the code were to be refactored in the future in a
way that might allow a null variable.

mp911de:
> This is a common best practice to prevent potential NullPointerExceptions if the variable (token) were to be null.

We have proper indication of nullable parameters and whether collections contain nullable elements hence we don't need to address an aspect that isn't broken.

PR #4048, opened 2025-10-17
Use constant-first equals in QueryTransformers

namest504:
#4047

Refactors QueryTransformers to use constant-first equals() calls. This is a common best practice to prevent potential NullPointerExceptions if the variable (token) were to be null.

```java
class QueryTokens {
    // ...
    static final QueryToken TOKEN_COMMA = token(", ");
    static final QueryToken TOKEN_AS = expression("AS");
    // ...
}
```

Before:
```java
if (token.equals(TOKEN_AS)) {
     // ...
 }
```

After:
```java
if (TOKEN_AS.equals(token)) {
    // ...
}
```

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryTransformers.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryTransformers.java
@@ -52,16 +52,16 @@ class QueryTransformers {
 					continue;
 				}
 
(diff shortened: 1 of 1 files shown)


mp911de:
> This is a common best practice to prevent potential NullPointerExceptions if the variable (token) were to be null.

We have proper indication of nullable parameters and whether collections contain nullable elements hence we don't need to address an aspect that isn't broken.

PR #4069, opened 2025-11-06
Avoid invoking callbacks in fallback EntityManagerFactory creation in AOT.

odrotbohm:
This is to avoid AOT repository generation failures in cases that require advanced customization of the EntityManagerFactory, such as Spring-provided entity listeners.


mp911de:
Thank you for your contribution. That's merged now.

PR #4071, opened 2025-11-07
Avoid repeated creation of `EntityManagerFactory` in AOT repository generation

### LINKED ISSUE #4068: Avoid repeated creation of `EntityManagerFactory` in AOT repository generation.
In AOT processing of the repositories, a fallback `EntityManagerFactory` is created to obtain information about the JPA model. That factory seems to be unnecessarily created once per repository.

mp911de:
Closes #4068


odrotbohm:
I've applied the patch locally and see the repeated lookups gone from the logs. 👍

PR #4108, opened 2025-11-29
Add private constructor and null validation to CollectionUtils.

qkrtkdwns3410:
## Summary

Improve `CollectionUtils` utility class with defensive programming practices.

### 1. Add private constructor
- Prevent instantiation of utility class
- Follow Java best practices for utility classes

### 2. Add null validation with `Assert.notNull()`
- Enforce the contract documented in Javadoc (`@param list must not be null`)
- Fail fast with clear error message instead of `NullPointerException`

### 3. Add unit tests for null validation
- `getFirstShouldRejectNullList()` 
- `getLastShouldRejectNullList()`

## Why these changes?

| Before | After |
|--------|-------|
| No instantiation prevention | Private constructor |
| Silent NPE on null input | Clear `IllegalArgumentException` |
| No test for null case | Test coverage added |

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/CollectionUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/CollectionUtils.java
@@ -17,12 +17,19 @@ package org.springframework.data.jpa.repository.query;
 
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


mp911de:
`CollectionUtils` is internal API, we don't need to add additional complexity and guards since this code is used internally only.

PR #4122, opened 2025-12-17
Fix grammar in projections documentation.

GoToBILL:
Remove unnecessary comma after 'both' in the sentence structure 'both A and B'. The comma between 'both' and the first item is grammatically incorrect.

  **Before:**
  Support for string-based queries covers both, JPQL queries(@Query) and native queries (@NativeQuery).

  **After:**
  Support for string-based queries covers both JPQL queries (@Query) and native queries (@NativeQuery).


schauder:
Thanks for the fix.

PR #4129, opened 2025-12-19
WIP: Prevent Hibernate basics types to be treated as DTO projection

### LINKED ISSUE #3929: Returning Converted entity properties cause DTO projection rewrite
When a Spring Data JPA query returns a view (interface or just a single property), it is now required that converted properties provide a redundant constructor with their own class as parameter. E.g.

```java
public interface DemoRepository extends JpaRepository<DemoEntity, Long> {
    @Query("SELECT e.valueObject1 FROM DemoEntity e")
    DemoValueObject1 findFirstValueObject1();
}
```
the `DemoValueObject1` (mapped by an `AttributeConverter` to string) needs such a really redundant ctor:
```
public class DemoValueObject1 {

    private String value;

    public DemoValueObject1(DemoValueObject1 value) {
        this.value = value.getValue();
    }
}
```
otherwise it fails with 
`org.hibernate.query.SemanticException: Missing constructor for type 'DemoValueObject2' [SELECT new com.example.demo.DemoValueObject2(e.valueObject2) FROM DemoEntity e]`

Here is the full minimal sample application to reproduce it:
https://github.com/hirth-abi/spring_boot_353_2/blob/main/src/test/java/com/example/demo/DemoRepositoryTest.java

Tested with Spring Boot 3.5.3, this was not required in Spring Boot 3.4.

joshiste:
fixes #3929

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/AbstractStringBasedJpaQuery.java
@@ -156,7 +156,7 @@ abstract class AbstractStringBasedJpaQuery extends AbstractJpaQuery {
 		}
 
 		if ((known != null && !known) || returnedJavaType.isArray() || getMetamodel().isJpaManaged(returnedJavaType)
-				|| !returnedType.needsCustomConstruction()) {
+				|| getMetamodel().isBasicType(returnedJavaType) || !returnedType.needsCustomConstruction()) {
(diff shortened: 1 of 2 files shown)


mp911de:
Thanks a lot. That looks like a decent addition. Care to add a test along with your changes?

joshiste:
@mp911de revisting this approach after the two weeks two things come to my mind: This might not take `AttributeConverter`, `UserTypes` or `@JavaType` annotations into account.
So I wonder wether it might be more helpful, to add a include/exclude/disable option for this feature?

PR #4132, opened 2025-12-22
Makes Nullability of findAll consistent.

### LINKED ISSUE #4131: Inconsistent `Specification` nullness declarations in `SimpleJpaRepository` and `JpaSpecificationExecutor`
In SimpleJpaRepository method getQuery contains contradiction:

```java
private <S extends T> TypedQuery<S> getQuery(ReturnedType returnedType, @Nullable Specification<S> spec,
			Class<S> domainClass, Sort sort, Collection<String> inputProperties, @Nullable ScrollPosition scrollPosition) {

		Assert.notNull(spec, "Specification must not be null");
```

Parameter "spec" is declared @Nullable but it's then immediately validated to be notNull

This additional check was introduced lately and turned out to be a breaking change

schauder:
All findAll methods now require a non null  except one, matching the current implementation and documentation in the method comments.
Changing it to non-null would have been a breaking change since the implemented interface defines the Specificaton as nullable.

Of course, since there is a null check in place any actual call with a null parameter would already fail, and therefore we might accept this breakage.

Of course again, changing the methods to a non-null argument is technically also a breaking change since they might have been overwritten.

Closes #4131

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
@@ -333,13 +333,13 @@ public class SimpleJpaRepository<T, ID> implements JpaRepositoryImplementation<T
 		Class<T> domainType = getDomainClass();
 
 		if (metadata == null) {
-			return Optional.of(entityManager.find(domainType, id));
+			return Optional.ofNullable(entityManager.find(domainType, id));
 		}
 
 		LockModeType type = metadata.getLockModeType();
 		Map<String, Object> hints = getHints();
 
-		return Optional.of(
+		return Optional.ofNullable(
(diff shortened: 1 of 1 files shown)


spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java
```diff
@@ -334,13 +333,13 @@ public Optional<T> findById(ID id) {
 		Class<T> domainType = getDomainClass();
 
 		if (metadata == null) {
-			return Optional.ofNullable(entityManager.find(domainType, id));
+			return Optional.of(entityManager.find(domainType, id));
```
mp911de: `EntityManager.find(…)` can return `null`.

schauder:
Ready for the next round.

mp911de:

Issue #4037, opened 2025-10-07
Refactor `JpaEntityInformation` handling in `repository.query` and `repository.support` packages


mp911de:
`JpaEntityInformation` is a central piece of information that we use across our repository and query implementations. We have to determine consistently an entity name that might come from XML (and not only `@Entity`), determine identifier columns and obtain the keyset for a given object.

Right now, we have several tangles and inconsistencies (e.g. if a domain class is not an `EntityType`, usages from `JpaKeysetScrollQueryCreator` in the `query` package). Ideally, all related information would reside inside `JpaPersistentEntity` or a similar type that we can put next to `JpaPersistentEntity` to unify identifier access (`JpaPersistentEntityImpl` and `JpaMetamodelEntityInformation` deal with identifiers and both in different ways).

mp911de:
Sure, feel free to give it a spin. Ideally, we provide a utility first that hosts identifier resolution and that can be used by `JpaPersistentEntity` and its accessor as well as from `JpaEntityInformation`. Both have slightly different needs for identifier access originating in the purpose. `JpaEntityInformation` is used for direct `EntityManager` interaction while `JpaPersistentEntity` is used in providing identifier details towards integration (e.g. Spring Data REST)/application code.

namgigun:
@mp911de Hi I'm interested in this issue. would it be okay to join the effort?

PR #4149, opened 2026-01-08
Introduce JpaIdentifierResolver and refactor identifier access

oTAETAEo:
- [x] You added yourself as author in the headers of the classes you touched. Amend the
date range in the Apache license header if needed. For new types, add the license header
(copy from another file and set the current year only).

## Motivation

  Centralize identifier resolution to remove scattered logic and keep behavior consistent across metadata/access paths, especially for composite ids.

  ## Summary

  - Introduce JpaIdentifierResolver to centralize identifier metadata/value access.
  - Delegate identifier access from JpaMetamodelEntityInformation and proxy-aware access from JpaPersistentEntityImpl.
  - Add unit tests covering single/composite ids, keyset handling, and non-entity projections.

  ## Tests

  - ./mvnw -pl spring-data-jpa -Dtest=JpaIdentifierResolverUnitTests test
  - ./mvnw -pl spring-data-jpa -Dtest=UserRepositoryTests#scrollByPredicateKeysetWithDtoProjection test
  - ./mvnw -pl spring-data-jpa -Dtest='JpaMetamodelEntityInformation*' -Dscan=false -Ddevelocity.storage.directory=/tmp/develocity-root
    (fails due to the EclipseLink javaagent issue described below)

  ## Notes / Question

  EclipseLink test discovery fails when the EclipseLink javaagent is enabled.

  java.lang.VerifyError: Bad type on operand stack
  (Type 'java.util.UUID' is not assignable to 'java.lang.String')

  When I removed the EclipseLink javaagent from the eclipselink-test argLine in spring-data-jpa/pom.xml, the test passed, which points to weaving as the trigger. I reverted that local
  change and would appreciate guidance on the preferred fix.

  Refs #4037

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/mapping/JpaIdentifierResolver.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/mapping/JpaIdentifierResolver.java
@@ -137,9 +137,18 @@ public final class JpaIdentifierResolver<T> {
 				return t.get(idMetadata.getSimpleIdAttribute().getName());
 			}
 
-			if (idMetadata.getJavaType().isInstance(entity)) {
+			Class<?> entityType = idMetadata.getJavaType();
+			if (entityType != null && entityType.isInstance(entity)) {
 				return persistenceUnitUtil.getIdentifier(entity);
 			}
+
+			return new DirectFieldAccessFallbackBeanWrapper(entity)
+					.getPropertyValue(idMetadata.getSimpleIdAttribute().getName());
+		}
+
+		Class<?> entityType = idMetadata.getJavaType();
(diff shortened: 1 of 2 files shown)


mp911de:
The design goes into a good direction but the change merely moves code around without addressing duplications (e.g. repetitive `DirectFieldAccessFallbackBeanWrapper` construction/extraction). 

Generating code isn't the problem, everyone of us is able to drop a few prompts into AI and come up with something. Refining the design, reducing complexity and  analyzing why tests fail (and fixing them) is the difficult part. 

I think you will have to spend a few more cycles to improve the overall design and unify the various ways of identifier extraction, understanding each variant and build an understanding whether these are still required or which parts can be merged as they share the same underlying concept.

PR #4154, opened 2026-01-12
Fix a/an in Javadoc

ngocnhan-tran1996:


schauder:
Thanks, that's merged and ported back.

PR #4156, opened 2026-01-13
Fix keyset pagination with hierarchical idclass

JBodkin-Amphora:
I've noted a few issues with keyset pagination when using hierarchical idclasses that I believe was introduced when Jakarta Persistence changed the `greaterThan`/`lessThan` methods on the criteriabuilder to take a Comparable. I've written some code to extract the attribute paths so that the underlying values can be extracted instead of the entity.

Whilst writing tests for the code, I found that the `isNotNull`/`isNull` predicates are being applied the wrong way around. This is covered by the tests.

I've noted that the Eclipse Link tests are now failing, I'm not sure if the technology supports the same hierarchical classes as I'm a Hibernate user. If it doesn't, is there a way to prevent the new repositories/entities being used for those tests?

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JpaMetamodelEntityInformation.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/JpaMetamodelEntityInformation.java
@@ -46,6 +46,7 @@ import org.springframework.data.jpa.repository.query.JpaMetamodelEntityMetadata;
 import org.springframework.data.jpa.util.JpaMetamodel;
 import org.springframework.data.util.DirectFieldAccessFallbackBeanWrapper;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 
 /**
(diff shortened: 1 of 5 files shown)


JBodkin-Amphora:
Eclipse Link tests are now passing after changes to the entities. Also changed the lookup to loop through the attributes instead of trying to look up by type, as Eclipse Link converts the primitive type to the wrapper type inside the metamodel

mp911de:
Thanks a lot. Which version of Spring Data JPA is affected for you? (So that we can correctly apply backports).

JBodkin-Amphora:
I've noted the issue starting when I migrated to Spring Boot v4, which uses the 4.0.x branch in this project.

JBodkin-Amphora:
I noted that the milestone for this has been closed, does this need a new milestone?

We've been running this internally for a while now and haven't come across any other issues.

mp911de:
We can sort this still out and apologies for the long radio silence. This ticket got burried under a ton of other tickets. I'll take a look next week.

mp911de:

Issue #4167, opened 2026-01-21
EclipseLink 5: derived query existsBy... fails due to implicit select alias


hgwr:
### Summary

After upgrading to Spring Boot 4.x (Spring Data JPA 4.0.x) with EclipseLink 5.0.0-B13, some derived queries fail at runtime.
This issue appears to be separate from #4145 (IN clause list binding).

### Background / Regression

We recently upgraded our project from Spring Boot 3.x to 4.x.
After the upgrade, several repository methods based on derived queries started failing at runtime unless rewritten using explicit `@Query` JPQL.

This suggests a regression or compatibility gap introduced by the combination of **Spring Data JPA 4.x** and **EclipseLink 5 (Hermes JPQL parser)**.

#### 1. `existsBy…` derived queries fail with JPQL syntax error

EclipseLink 5 (Hermes) rejects JPQL generated for `existsBy…` queries due to an implicit alias in the select clause.

Example exception:

```
BadJpqlGrammarException: ... Syntax error parsing ... select o.id id ...
```

Generated JPQL (example):

```sql
SELECT o.id id FROM OfUser o
WHERE UPPER(o.email) = UPPER(?1)
  AND o.id != ?2
  AND o.isDelete != ?3
```

Hermes appears to reject `select o.id id` (implicit alias without `AS`).

#### 2. Parameter mismatch in `findByIdAndIsDeleted`

Some derived queries fail with a positional parameter binding error:

```
InvalidDataAccessApiUsageException:
Query argument 2 not found in the list of parameters provided during query execution
```

This seems to indicate that EclipseLink 5 under-reports positional parameters for certain derived queries.

### Expected behavior

Derived queries should execute successfully on EclipseLink 5 without requiring explicit JPQL annotations.

#### Entity

```java
@Entity
class SoftDeleteUser {

  @Id
  @GeneratedValue
  Long id;

  @Column(nullable = false)
  String email;

  @Column(nullable = false)
  Integer isDeleted;
}
```

#### Repository

```java
interface SoftDeleteUserRepository
    extends JpaRepository<SoftDeleteUser, Long> {

  boolean existsByEmailIgnoreCaseAndIdNotAndIsDeletedNot(
      String email, Long id, Integer isDeleted);

  Optional<SoftDeleteUser> findByIdAndIsDeleted(
      Long id, Integer isDeleted);
}
```

#### Test scenario

1. Persist two users with `isDeleted = 0`
2. Call
[...]

PR #4170, opened 2026-01-23
Fix positional binding when only later positions are reported

hgwr:
Fix binding for EclipseLink when a derived query collapses to `... IS NULL AND ... = ?2` and the provider reports only position 2. We now bind positional parameters based on the reported positions (falling back to the previous size heuristic when no positional metadata is available).

This prevents `Query argument 2 not found` seen in #4167. Reproducer and logs: https://github.com/spring-projects/spring-data-jpa/issues/4167#issuecomment-3783485530.

Tests: ./mvnw -pl spring-data-jpa -Dtest=NamedOrIndexedQueryParameterSetterUnitTests test (local `eclipselink-test` fork still fails due to the existing EclipseLink javaagent ASM error).

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParameterSetter.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryParameterSetter.java
@@ -109,10 +109,12 @@ interface QueryParameterSetter {
 			} else {
 
 				Integer position = parameter.getPosition();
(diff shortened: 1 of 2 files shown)


mp911de:
The problem is that the parameter binding index is being incremented leading to gaps in numbering and that is the actual problem to be fixed. Closing in favor of 747a3ee14d4f5ad70219ed0aa835aa7ae1763300.

PR #4183, opened 2026-02-10
Refactor entity metadata resolution to use JpaPersistentEntity.

namgigun:
1. Refactor JpaPersistentEntityImpl.getIdentifier to use PersistenceProvider instead of ProxyIdAccessor for proxy identification checks.
2. Introduce JpaPersistentEntityInformation to decouple metadata extraction from JpaMetamodelEntityInformation. This allows metadata resolution by depending solely on JpaPersistentEntity rather than multiple instances like IdMetadata or PersistenceUnitUtil.
3. Add JpaPersistentEntityInformationUnitTests to verify identity extraction, type resolution, and composite ID detection logic.

This change aligns JPA entity information handling with the core Spring Data mapping infrastructure and reduces redundant metadata analysis. Refs #4037


namgigun:
Could you please confirm whether the required code formatters have been properly applied to this change?

PR #4218, opened 2026-03-19
Refine `JpaOrder.withUnsafe(…)` documentation

### LINKED ISSUE #2932: JpaSort.JpaOrder.withUnsafe(String ... properties) is somewhat unintuitive.
We should at the very least write some tests to document the intended behaviour and also clarify that in the documentation.

soowanx:
Closes #2932

This clarifies that `JpaOrder.withUnsafe` returns a new order instance instead of mutating the existing one.

The change adds focused tests and updates the reference documentation to make that behavior explicit.

Tests:
- `JpaSortTests`


mp911de:

PR #4249, opened 2026-05-07
Detect `COUNT(*)` function alias in `QueryUtils`

### LINKED ISSUE #4242: `COUNT(*) AS alias` is not detected by `QueryUtils FUNCTION_PATTERN`
### Summary

  `QueryUtils.FUNCTION_PATTERN` (QueryUtils.java:187) uses the character
  class `[\w\.,\s'=:;\\?]+` for function arguments. The class does not
  include `*`, so the extremely common `COUNT(*) AS <alias>` form is not
  detected by `getFunctionAliases(…)`.

  When `DefaultQueryEnhancer` is the active enhancer (native queries,
  default setup without the optional `jsqlparser` dependency),
  `applySorting(…)` then prefixes the alias with the root entity alias,
  producing invalid SQL like `ORDER BY u.cnt` instead of `ORDER BY cnt`.

  ### Reproducer

  ```java
  @Test
  void functionAliasPatternShouldDetectCountStar() {
      assertThat(QueryUtils.getFunctionAliases("SELECT COUNT(*) as total FROM User u"))
              .containsExactly("total");
  }
  ```
  Current result: empty set. Existing tests in QueryUtilsUnitTests
  (around line 550) cover COUNT(1) but never COUNT(*).

  Suggested fix

  // QueryUtils.java:187
  builder.append("\\w+\\s*\\([\\w\\.,\\s'=:;\\\\?\\*]+\\)");

  Related

  I am aware of #2322 and #2441, which reported broader regex limitations
  (nested parens, complex casts) and were closed as feedback-provided.
  This issue is deliberately narrower: the one-character fix above
  addresses the specific COUNT(*) case without attempting to solve the
  general parsing problem.

  ---

config25:
Include the asterisk character in the `FUNCTION_PATTERN` argument
  character class so that the `COUNT(*) AS <alias>` form is detected by
  `QueryUtils.getFunctionAliases`.

  Without this, `DefaultQueryEnhancer.applySorting` prefixes the alias
  with the root entity alias and produces invalid SQL such as
  `ORDER BY u.cnt` instead of `ORDER BY cnt`.

  Adds a regression test (`discoversCountStarFunctionAlias`) covering the
  issue's reproducer and the GROUP BY example from the issue body.

  Closes #4242

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryUtils.java
@@ -90,6 +90,7 @@ import org.springframework.util.StringUtils;
  * @author Yanming Zhou
  * @author Alim Naizabek
  * @author Jakub Soltys
+ * @author Young-ho Kim
  */
 public abstract class QueryUtils {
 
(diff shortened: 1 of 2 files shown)


mp911de:

PR #4255, opened 2026-05-25
Fix `EXTRACT(TIME FROM ...)` in `JpaSort` order expressions

oniwon:
`HqlOrderExpressionVisitor#getTemporalField(ExtractFieldContext)` returned `LocalDateTimeField.DATE` for the `TIME` field — a copy-paste from the `DATE` branch just above. As a result, sorting via `JpaSort.unsafe("extract(time from …)")` silently extracted the **date** portion instead of the **time** portion (wrong ordering, no exception). Introduced in b61c51d8c.

This PR returns `LocalDateTimeField.TIME` and adds regression coverage in `HqlOrderExpressionVisitorUnitTests#extract` for both the DATE and TIME extract fields (`extract(date …)` → `cast(… as LocalDate)`, `extract(time …)` → `cast(… as LocalTime)`).

- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/main/CONTRIBUTING.adoc).
- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/main/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlOrderExpressionVisitor.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlOrderExpressionVisitor.java
@@ -60,6 +60,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Greg Turnquist
  * @author Mark Paluch
+ * @author oniwon
  * @since 4.0
  */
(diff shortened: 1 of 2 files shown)


mp911de:

PR #4262, opened 2026-05-29
Document fetch behavior for streaming query results

### LINKED ISSUE #1346: Add note about driver behaviour for streaming query results [DATAJPA-1007]
**[Robert Hunt](https://jira.spring.io/secure/ViewProfile.jspa?name=roberthunt87@gmail.com)** opened **[DATAJPA-1007](https://jira.spring.io/browse/DATAJPA-1007?redirect=false)** and commented

It would be useful to add a note about JDBC driver behaviour in relation to streaming results. I know it goes beyond the scope/control of Spring Data but someone might assume that by using a Stream return type that the ResultSet is being fetched in a streaming nature from the database.

In the case of MySQL this is certainly not the case as a `@QueryHint` needs to be included as a signal to the underlying driver (see: http://knes1.github.io/blog/2015/2015-10-19-streaming-mysql-results-using-java8-streams-and-spring-data.html)

Just a small note as a hint to the reader that they may need to investigate the behaviour of their specific drivers in regard to memory-efficient fetching would be useful

---

**Reference URL:** http://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-streaming

jewoodev:
Adds a JPA-specific note to the Query Hints section for repository methods returning `Stream<T>`.

A `Stream<T>` return type can suggest lazy result processing, but the actual row fetching behavior still depends on the JPA provider and JDBC driver. The note clarifies that returning a `Stream<T>` does not guarantee incremental row fetching. It places the guidance alongside `@QueryHints`, since query hints are the JPA mechanism for applying provider-specific fetch-size settings.

Testing: `./mvnw -pl spring-data-jpa-distribution -am -Pantora install -DskipTests`

Closes #1346

### DIFF
--- a/src/main/antora/modules/ROOT/pages/jpa/query-methods.adoc
+++ b/src/main/antora/modules/ROOT/pages/jpa/query-methods.adoc
@@ -684,6 +684,9 @@ public interface UserRepository extends Repository<User, Long> {
 ====
 The preceding declaration would apply the configured `@QueryHint` for that actually query but omit applying it to the count query triggered to calculate the total number of pages.
 
+NOTE: Returning a `Stream<T>` does not guarantee that the JPA provider or JDBC driver fetches database rows incrementally.
+For memory-efficient processing of large result sets, use query hints to apply provider-specific fetch-size settings and check your JPA provider and JDBC driver documentation for supported values.
+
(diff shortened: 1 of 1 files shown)


mp911de:

PR #4273, opened 2026-06-04
Fix JPQL, EQL, and HQL query rendering round-trips

### LINKED ISSUE #4272: JPQL and EQL renderers fail using `TREAT`, `LOCATE`, and subqueries from collection members
`JpqlQueryRenderer` and `EqlQueryRenderer` are expected to re-render a parsed query unchanged, but three Jakarta Persistence spec-valid constructs are currently dropped or fail during rendering. All three parse fine in JPQL/EQL, and the same round-trip cases pass through the HQL renderer, so the failures are isolated to the JPQL/EQL visitors.

| Construct | Spec-valid input (abbrev.) | Current behavior |
|---|---|---|
| `TREAT(…)` downcast in a path | `SELECT TREAT(VALUE(m) AS Employee) …` | `SELECT VALUE(m) …` — cast silently dropped |
| Two-argument `LOCATE(…)` | `LOCATE('a', e.name)` | `NullPointerException` while rendering |
| Subquery `FROM` collection member | `… FROM Order o2, IN(o2.lineItems) l …` | `… FROM Order o2 …` — member dropped, alias `l` left undefined |

Each is independently reproducible against the current renderers. I can submit a focused PR with one commit per fix.

jewoodev:
These four focused commits fix renderer-only defects where the JPQL, EQL, and HQL renderers fail to preserve parsed query constructs during rendering. The first three affect `JpqlQueryRenderer` and `EqlQueryRenderer` (HQL already round-trips these correctly); the fourth fixes the HQL-only `column(...)` function in `HqlQueryRenderer`. All four are confined to visitor methods, add round-trip tests, and leave the grammars untouched.

| Fix | Root cause | Change |
|---|---|---|
| `TREAT(…)` downcast | `TREAT` branch guarded by a condition duplicated from the preceding branch → unreachable, cast silently dropped | check `TREAT()` first |
| Two-argument `LOCATE(…)` | third-argument guard used `!= null` on a never-null list accessor → `visit(null)`, throwing `NullPointerException` | check the list is non-empty |
| Subquery `FROM` collection member | `visitSubquery_from_clause` rendered only the identification variable, dropping a trailing `collection_member_declaration` and leaving its alias undefined | iterate child nodes in source order |
| HQL `column(…)` cast target | `visitColumnFunction` appended the function name as an expression after the `.` (stray space) and, for the optional cast, visited `jpaNonstandardFunctionName()` instead of `castTarget()` → `column(tbl.foo as int)` rendered as `column(tbl. foo as foo)` | append without expression spacing and render `castTarget()`, matching `visitJpaNonstandardFunction` |

Closes #4272

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EqlQueryRenderer.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/EqlQueryRenderer.java
@@ -554,8 +554,15 @@ class EqlQueryRenderer extends EqlBaseVisitor<QueryTokenStream> {
 		QueryRendererBuilder builder = QueryRenderer.builder();
 
 		builder.append(QueryTokens.expression(ctx.FROM()));
-		builder.appendExpression(
-				QueryTokenStream.concat(ctx.subselect_identification_variable_declaration(), this::visit, TOKEN_COMMA));
+
+		List<ParseTree> declarations = new ArrayList<>();
+		for (ParseTree child : ctx.children) {
+			if (child instanceof EqlParser.Subselect_identification_variable_declarationContext
+					|| child instanceof EqlParser.Collection_member_declarationContext) {
+				declarations.add(child);
+			}
+		}
+		builder.appendExpression(QueryTokenStream.concat(declarations, this::visit, TOKEN_COMMA));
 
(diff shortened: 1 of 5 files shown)


mp911de:

PR #4274, opened 2026-06-05
Remove unused QueryRendererBuilder locals in JPQL and HQL query renderers

jewoodev:
While following up on #4273, I found two leftover `QueryRendererBuilder` locals that are never used.

This removes the unused builder in `JpqlQueryRenderer#visitString_expression` and the empty `nested` builder plus no-op `appendInline(nested)` in `HqlQueryRenderer#visitListaggFunction`. Rendering behavior is unchanged.

Verified with the JPQL/HQL/EQL renderer round-trip suites.

- [ ] You submit test cases (unit or integration tests) that back your changes. — N/A for this behavior-preserving dead-code removal; the existing renderer round-trip suites noted above already exercise both methods and pass unchanged.


mp911de:

PR #4275, opened 2026-06-05
Fix `IS NULL` rendering of keyset queries using Querydsl

stupid58fly:
This PR fixes the issue for querydsl scroll where one of keyset position values is null.
Before fix querydsl generates query with `column = null` and after fix it will be `column is null`.


mp911de:
Can we have simpler tests, ideally that demonstrate the problem and test the query itself rather than the effect?

stupid58fly:
> Can we have simpler tests, ideally that demonstrate the problem and test the query itself rather than the effect?

I tried to write this test, but didn't succeed.
There is no such tests for QueryDSL, at least I could not find it.
I have tried to do that, but couldn't find any point to join the query rendering.
Not sure it is possible to implement such test without refactoring QueryDSL implementation.
If you have any suggestions, could you please share?

mp911de:

PR #4276, opened 2026-06-06
Add regression tests for sorting native queries with a derived table containing `ORDER BY … LIMIT`

### LINKED ISSUE #2395: Pageable - incorrect query for complex join for  nativeQuery=true
Hello,

spring-data-jpa, version: **2.3.9**
database: **mysql**

An error occures for method annotated with `@Query(nativeQuery=true)` with declared _pageable_ parameter, e.g.:

`Page<NEntity> searchForEntityWithStatus(@Param("status"), Pageable pageable);`

If query contains complex JOIN with ORDER BY and LIMIT, e.g.:

`"LEFT JOIN( SELECT status FROM entity_statuses ORDER BY updated_at  DESC LIMIT 1) st ON ..."`

library generates wrong sorting and pagination  fragment (_ORDER BY_ is missing):

 `, some_column LIMIT ?,?` which causes SQL error.

Workaroundthat working for me is to add at the end of query sentence like below:

`ORDER BY <any_neutral_column>`

_any_neutral_column_ I mean column which has no impact on sorting.

jewoodev:
The broken sort fragment reported in #2395 reproduces on the version it was filed against (2.3.9): `QueryUtils.applySorting` matched the derived table's inner `ORDER BY` and appended the primary sort after the `WHERE` clause without an `order by` keyword (`… = :status, n.name asc`), which becomes invalid SQL once pagination adds its `LIMIT`. It is no longer reproducible on `main` — both native query enhancers append the primary `ORDER BY` to the outer query while preserving the derived table's inner ordering.

This PR adds a regression test for each native enhancer: the regex-based `DefaultQueryEnhancer`, where the original breakage lived, and the JSqlParser-based `JSqlParserQueryEnhancer`, the default once JSqlParser is on the classpath. Each sits next to the existing `shouldApplySorting` coverage in its test class; both classes pass locally.

Closes #2395

---

### DIFF
--- a/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/DefaultQueryEnhancerUnitTests.java
+++ b/spring-data-jpa/src/test/java/org/springframework/data/jpa/repository/query/DefaultQueryEnhancerUnitTests.java
@@ -29,6 +29,7 @@ import org.springframework.data.repository.query.ReturnedType;
  *
  * @author Mark Paluch
  * @author Alim Naizabek
+ * @author Jewoo Shin
  */
 class DefaultQueryEnhancerUnitTests extends QueryEnhancerTckTests {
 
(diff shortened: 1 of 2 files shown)


jewoodev:
Closing this — on a closer look it doesn't add meaningful coverage.

The behaviour #2395 hinges on (`QueryUtils.applySorting` keeping the outer `ORDER BY` when a `FROM`-clause derived table carries its own `ORDER BY … LIMIT`) is already exercised by `QueryUtilsUnitTests`, and the issue no longer reproduces on `main`. A test-only PR here would just duplicate that existing coverage.

Issue #4278, opened 2026-06-09
JPQL and EQL fail to parse a single-character string literal as the left operand of `MEMBER OF`


jewoodev:
JPQL and EQL currently fail to parse a single-character quoted string when it is used as the left operand of `MEMBER OF`, throwing a `BadJpqlGrammarException`. The multi-character form parses fine, and the same query parses through the HQL grammar, so for this input the failure is isolated to the JPQL and EQL grammars.

```sql
-- fails with BadJpqlGrammarException: no viable alternative at input "'c' *MEMBER OF ..."
SELECT e FROM Employee e WHERE 'c' MEMBER OF e.responsibilities

-- parses fine (multi-character literal)
SELECT e FROM Employee e WHERE 'write code' MEMBER OF e.responsibilities
```

Root cause: the lexer tokenizes a single-quoted value such as `'c'` as `CHARACTER`, while values such as `'write code'` and `''` are tokenized as `STRINGLITERAL`. The parser rule `literal` lists `STRINGLITERAL` but omits `CHARACTER`, whereas its sibling rule `string_literal` lists both. Among the direct uses of `literal`, this surfaces on the `MEMBER OF` left operand: the `IN` and constructor paths can also reach `string_literal` through `string_expression`, so their single-character cases still parse.

Jakarta Persistence describes string literals as single-quoted strings and does not distinguish one-character strings from longer strings; the `CHARACTER`/`STRINGLITERAL` split is an implementation detail of these grammars. I can submit a focused PR adding `CHARACTER` to the `literal` rule in `Jpql.g4` and `Eql.g4` (mirroring `string_literal`), covered by a TCK round-trip test.

PR #4282, opened 2026-06-10
Fix parsing of single-character string literals in JPQL and EQL collection member expressions.

jewoodev:
JPQL and EQL currently tokenize a single-character quoted string such as `'c'` as `CHARACTER`, while other quoted strings are tokenized as `STRINGLITERAL`. The `literal` parser rule accepted `STRINGLITERAL` but not `CHARACTER`, causing a single-character string literal to fail when used as the left operand of `MEMBER OF`.

This change adds `CHARACTER` to the JPQL and EQL `literal` rules and covers the collection-member case with a shared query renderer round-trip test.

Jakarta Persistence allows a `literal` as the left operand of `MEMBER OF`, defines string literals as single-quoted strings without distinguishing single-character strings from longer strings, and itself uses `'Joe' MEMBER OF p.nicknames` as an example.

See #4278

### DIFF
--- a/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
+++ b/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 the original author or authors.
+ * Copyright 2023-present the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ grammar Eql;
  * @see https://eclipse.dev/eclipselink/documentation/3.0/jpa/extensions/jpql.htm
  * @author Greg Turnquist
  * @author Christoph Strobl
+ * @author Jewoo Shin
  * @since 3.2
  */
 }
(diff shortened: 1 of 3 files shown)


mp911de:
Thanks a lot. Do you have a few more examples where the single-char case falls short? Also, getting an overview of unintentionally enabling functionality that wasn't previously possible (or breaking existing) would be great. We might want to introduce a bit more test coverage and cross-check whether we have similar issues in the HQL parser.

jewoodev:
Thanks for the feedback.

I pushed additional TCK coverage for the same root cause in collection-member expressions: `NOT MEMBER OF` and the same `MEMBER OF` predicate inside subqueries, `UPDATE` where clauses, and `DELETE` where clauses.

I also found `MEMBER` without `OF` while probing. These cases fail in JPQL and EQL on `main` and pass with this branch. Other common single-character string literal positions such as comparisons, `IN` items, constructor arguments, `LIKE`, `TRIM`, `FUNCTION`, `CASE`, and `UPDATE` assignments already parse on `main` because they can route through `string_expression`/`string_literal` rather than `literal`.

I will follow up with the check for unintended enablement or breakage, and with the HQL parser cross-check.

mp911de:
Alright, let me merge this PR. Once there is more, we can commit changes against the original bug report #4278

mp911de:

PR #4285, opened 2026-06-16
Add `EntityGraphHint` query method parameter

### LINKED ISSUE #4175: Consider EntityGraph argument for query methods
It would be neat to provide an entity graph to a query method through a Spring Data JPA type allowing:

* Declaration through typed property paths (e.g. `EntityGraphHint.fetch(Person::getFirstName, Person::getAddress, …)`)
* From a named entity graph reference (`EntityGraphHint.fetch("my-named-graph")`, `EntityGraphHint.load("my-named-graph")`)
* From a `EntityGraph` object (`EntityGraphHint.fetch(obj)`, `EntityGraphHint.load(obj)`)

Within this ticket, we should also consider the following changes:

- [ ] https://github.com/spring-projects/spring-data-jpa/issues/2113
- [ ] https://github.com/spring-projects/spring-data-jpa/issues/1746
- [ ] https://github.com/spring-projects/spring-data-jpa/issues/1334

yeomin4242:
Repository query methods can now accept `EntityGraphHint` to select an entity graph per invocation.

The hint can be created from type-safe property paths, named entity graphs, or a JPA `EntityGraph` instance. We treat the hint as a special parameter so it is excluded from query binding and apply it for runtime and AOT query creation. If the argument is `null`, query creation falls back to the declared `@EntityGraph` metadata.

Includes unit and integration tests for parameter discovery, runtime query creation, repository execution, and AOT query generation.

Closes #4175

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/EntityGraphHint.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/EntityGraphHint.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.jpa.repository;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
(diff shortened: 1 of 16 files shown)


jewoodev:
One precedence case seems worth pinning down with a regression test: if a query method declares `@EntityGraph` and receives a non-null `EntityGraphHint`, the parameter hint should override the declared graph.

The implementation currently does that, while the tests cover the hint-only path and the `null` fallback path. Adding a runtime/AOT test for the non-null override case would make the new invocation-level contract explicit.

yeomin4242:
> One precedence case seems worth pinning down with a regression test: if a query method declares `@EntityGraph` and receives a non-null `EntityGraphHint`, the parameter hint should override the declared graph.
> 
> The implementation currently does that, while the tests cover the hint-only path and the `null` fallback path. Adding a runtime/AOT test for the non-null override case would make the new invocation-level contract explicit.

@jewoodev Thanks!

I added regression tests for this in both the runtime and AOT paths. The tests invoke a method that declares `@EntityGraph` with a non-null `EntityGraphHint` for a different association, then assert that the parameter hint takes effect and the declared graph does not.

That should make the intended precedence explicit: non-null `EntityGraphHint` overrides the declared graph, while `null` continues to fall back to `@EntityGraph`.

PR #4289, opened 2026-06-23
Add HQL single-character member coverage.

jewoodev:
Add focused HQL renderer coverage for single-character string literals used as the left operand of collection membership predicates.

As a follow-up to #4278, I also checked the JPQL/EQL grammar change for unintended widening. The only meaningful new parse surface is the intended collection membership predicate shape (`MEMBER OF`/`NOT MEMBER OF`); common existing single-character literal positions such as comparisons, `IN`, constructors, `LIKE`, `TRIM`, functions, `CASE`, and update assignments continue to parse as before.

HQL does not share the same root cause because quoted strings are handled as `STRING_LITERAL`, without a separate single-character token. This PR leaves the HQL grammar unchanged and adds coverage for `member of`, `not member of`, subquery, update, and delete surfaces.

Verified with the HQL query renderer test suite.

See #4278


mp911de:

PR #4292, opened 2026-06-29
Allow KEY and VALUE in JPQL/EQL collection membership predicates.

jewoodev:
Allow JPQL and EQL collection membership predicates to use map key and value expressions as the left operand.

Jakarta Persistence section 4.6.9 describes the left operand with `state_valued_path_expression`, which includes map field identification variables such as `KEY(...)` and `VALUE(...)`. The current JPQL/EQL grammars use the narrower `state_field_path_expression` at this point, so queries such as `SELECT i FROM Item i JOIN i.photos p WHERE VALUE(p) MEMBER OF i.photoValues` fail before rendering.

HQL already accepts the same shape, and Hibernate ORM and EclipseLink accept these forms at `EntityManager#createQuery(...)` time.

This PR adds `map_field_identification_variable` to the JPQL/EQL collection membership left operand and covers `VALUE(...) MEMBER OF`, `KEY(...) MEMBER OF`, and `VALUE(...) NOT MEMBER OF` in the shared renderer TCK.

Verified with ANTLR generation and the JPQL/EQL/HQL query renderer test suite.

### DIFF
--- a/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
+++ b/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
@@ -388,6 +388,7 @@ collection_member_expression
 entity_or_value_expression
     : single_valued_object_path_expression
     | state_field_path_expression
+    | map_field_identification_variable
     | simple_entity_or_value_expression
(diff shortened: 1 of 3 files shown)


mp911de:

Issue #1731, opened 2018-09-05
Number of Params in complex Native Query is wrong [DATAJPA-1417]


spring-projects-issues:
**[Thomas Laegler](https://jira.spring.io/secure/ViewProfile.jspa?name=thlaegler)** opened **[DATAJPA-1417](https://jira.spring.io/browse/DATAJPA-1417?redirect=false)** and commented

Dear Spring Data JPA Team,

I have a quite complex native Query (PostgreSQL with PostGIS) with 3 Parameters (tripId, lat, lng). But it looks like Spring is only detecting 2 Parameters in the Query.

The repository looks like this:

```java
@SqlResultSetMapping(name = "StopDistanceMapping", entities = {@EntityResult(entityClass = StopDistance.class, fields = {@FieldResult(name = "stopId", column = "stop_id"), @FieldResult(name = "code", column = "stop_code"), @FieldResult(name = "lat", column = "stop_lat"), @FieldResult(name = "lng", column = "stop_lon"), @FieldResult(name = "sequence", column = "stop_sequence"), @FieldResult(name = "distance", column = "distance")})})
[...]

PR #4293, opened 2026-06-30
Reject missing explicitly named queries

014-code:
DATAJPA-1417 changes named query lookup so that repository methods explicitly declaring `@Query(name = …)` fail fast when the named query cannot be resolved.

Previously, if the configured named query was missing from both `NamedQueries` and the `EntityManager`, lookup could continue into query derivation. That fallback produced a misleading parameter-count error for methods whose name did not match the declared query parameters.

### Changes

- Reject unresolved explicit `@Query(name = …)` declarations with a `QueryCreationException`.
- Keep existing fallback behavior for methods that do not explicitly declare a named query.
- Add a regression test covering the missing explicit named query path.

### Tests

- `.\mvnw.cmd -pl spring-data-jpa -Dtest=JpaQueryLookupStrategyUnitTests test`
- `.\mvnw.cmd -pl spring-data-jpa -Dtest=PreprocessedQueryUnitTests test`

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryLookupStrategy.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaQueryLookupStrategy.java
@@ -171,7 +171,15 @@ public final class JpaQueryLookupStrategy {
 
 			RepositoryQuery query = NamedQuery.lookupFrom(method, em, configuration);
 
-			return query != null ? query : NO_QUERY;
+			if (query != null) {
+				return query;
+			}
+
+			if (method.hasAnnotatedQueryName()) {
+				throw QueryCreationException.create(method, String.format("Did not find named query '%s'", name));
+			}
+
+			return NO_QUERY;
 		}
 
 		private @Nullable DeclaredQuery getCountQuery(JpaQueryMethod method, NamedQueries namedQueries, EntityManager em) {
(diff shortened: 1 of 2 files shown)


mp911de:
I fail to understand how this pull request relates to #1731. Care to elaborate? The change on its seems reasonable.

014-code:
> I fail to understand how this pull request relates to #1731. Care to elaborate? The change on its seems reasonable.

Thanks for calling that out. The relation is in the lookup path shown by the stack trace rather than in native SQL parameter parsing itself.

In https://github.com/spring-projects/spring-data-jpa/issues/1731, the repository method declares only @Query(name = "StopDistanceQuery"). If that named query is not visible to the EntityManager, query lookup currently returns NO_QUERY and CREATE_IF_NOT_FOUND falls back to query derivation.

That fallback is visible in the reported stack trace:

CreateIfNotFoundQueryLookupStrategy -> PartTreeJpaQuery

The parameter-count error is therefore produced while deriving findNextStopByTripIdAndPosition, not while executing the named native query. The method has three bindable parameters, while the derived query path sees only two parts from the method name.

This PR changes that failure mode so an explicitly declared @Query(name = …) fails fast when the named query cannot be resolved, instead of falling back to query derivation and surfacing the misleading parameter-count error.

It does not attempt to fix the provider-level native query parameter parsing mentioned in the later update. If you consider that to be the actual scope of https://github.com/spring-projects/spring-data-jpa/issues/1731, I can remove the Closes #1731 reference and keep this as a related query lookup improvement.

mp911de:
Thanks for sharing your thoughts.

I removed the relationship between the ticket and this pull request so we can treat it as a separate item. Since the change is a clear improvement, we can handle it within the scope of this PR.

With this change, a misspelled query name would silently fall back to query derivation. That is a relevant behavioral change for existing applications that currently boot successfully and would fail after this change.

mp911de:

PR #4294, opened 2026-07-04
Allow whitespace in JPQL and EQL JDBC escape literals.

jewoodev:
JPQL and EQL tokenize JDBC date, time, and timestamp escape literals as single tokens, requiring the marker to be directly followed by the quoted literal. This rejects common JDBC-style forms such as `{d '2012-01-03'}`.

This change aligns JPQL and EQL parsing with HQL's `*_ESCAPE_START` structure. A shared JPQL, EQL, and HQL rendering path preserves both whitespace and no-whitespace forms, covered by the shared renderer TCK.

### DIFF
--- a/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
+++ b/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
@@ -699,9 +699,25 @@ pattern_value
 
 date_time_timestamp_literal
     : STRINGLITERAL
-    | DATELITERAL
-    | TIMELITERAL
-    | TIMESTAMPLITERAL
+    | jdbc_date_literal
+    | jdbc_time_literal
+    | jdbc_timestamp_literal
+    ;
+
+jdbc_date_literal
+    : DATE_ESCAPE_START generic_temporal_literal_text '}'
+    ;
+
+jdbc_time_literal
+    : TIME_ESCAPE_START generic_temporal_literal_text '}'
+    ;
+
+jdbc_timestamp_literal
+    : TIMESTAMP_ESCAPE_START generic_temporal_literal_text '}'
+    ;
+
+generic_temporal_literal_text
+    : STRINGLITERAL
     ;
 
 entity_type_literal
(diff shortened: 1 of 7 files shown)


spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
```diff
@@ -1030,6 +1030,6 @@ STRINGLITERAL               : '\'' (~ ('\'' | '\\')|'\\')* '\'' ;
 FLOATLITERAL                : ('0' .. '9')* '.' ('0' .. '9')+ (E ('0' .. '9')+)* (F|D)?;
 INTLITERAL                  : ('0' .. '9')+ ;
 LONGLITERAL                 : ('0' .. '9')+ L;
-DATELITERAL                 : '{' D STRINGLITERAL '}';
-TIMELITERAL                 : '{' T STRINGLITERAL '}';
-TIMESTAMPLITERAL            : '{' T S STRINGLITERAL '}';
+DATELITERAL                 : '{' D [ \t\r\n]* STRINGLITERAL '}';
```
mp911de: How about switching the general logic towards `TIMESTAMP_ESCAPE_START (dateTime | genericTemporalLiteralText) '}'` style as it is done in the HQL parser? The `TIMESTAMP_ESCAPE_START` is a bit strange though but it seems to require much less ceremony. What do you think?

jewoodev: Yes, I agree that this structure is cleaner. I updated JPQL and EQL to use `*_ESCAPE_START ... '}'` parser rules, following HQL's approach. Since JPQL and EQL do not have HQL's structured `dateTime`, `date`, and `time` rules, they retain their existing quoted-literal scope through `generic_temporal_literal_text`.

Splitting the composite lexer tokens also exposed a round-trip detail: the default renderer adds a space before `}`, while always rendering inline changes the existing no-whitespace forms. I therefore introduced a shared JDBC escape rendering path for JPQL, EQL, and HQL that preserves whether whitespace follows the marker. The shared renderer TCK now covers both forms for date, time, and timestamp literals.

PR #4295, opened 2026-07-06
Allow `REGEXP` modifier in HQL `LIKE` predicates.

jewoodev:
Hibernate's HQL parser accepts `LIKE REGEXP` and `ILIKE REGEXP` predicates, but Spring Data JPA's HQL grammar currently treats `regexp` as the pattern expression and rejects the actual pattern literal.

This change adds `REGEXP` as an HQL token and allows it as a modifier after `LIKE` or `ILIKE`, keeping standalone `REGEXP` predicates out of scope. HQL renderer round-trip coverage now verifies `LIKE REGEXP`, `NOT LIKE REGEXP`, and `ILIKE REGEXP`.

Verified with `./mvnw -pl spring-data-jpa generate-sources` and `./mvnw -pl spring-data-jpa -Dtest=HqlQueryRendererTests test`.

### DIFF
--- a/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Hql.g4
+++ b/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Hql.g4
@@ -1487,7 +1487,7 @@ betweenExpression
 
 // https://docs.jboss.org/hibernate/orm/6.1/userguide/html_single/Hibernate_User_Guide.html#hql-like-predicate
 stringPatternMatching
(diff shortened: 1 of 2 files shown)


stupid58fly:
Hi @jewoodev

Operators `(i?)like regexp` works for hibernate 7.2+
But did you check this expression for eclipse link?

Per my understanding, eclipse link supports `regexp` without `(i?)like`

jewoodev:
EQL parser already supports the standalone form `e.name REGEXP 'Jo.*'`.

This PR is scoped to Hibernate HQL, did you have a specific interaction with the EQL path in mind that you think this PR should address?

stupid58fly:
No, I just saw your changes and thought about a potential problem. 
I am very glad that I was wrong in my assumption.

mp911de:

Issue #4197, opened 2026-02-24
Explore support for JPA 4.0 with a JPA 3.2 baseline


mp911de:
To allow usage of early JPA 4.0 functionality, it would make sense to explore what is necessary to allow upgrading to Hibernate 8 as JPA 4.0 introduces several binary-incompatible changes such as: 

```
Caused by: java.lang.NoSuchMethodError: 'jakarta.persistence.Query jakarta.persistence.EntityManager.createNativeQuery(java.lang.String, java.lang.Class)'
    at org.springframework.data.jpa.repository.query.NativeJpaQuery.createJpaQuery(NativeJpaQuery.java:104) ~[spring-data-jpa-4.1.0-M1.jar:4.1.0-M1]
    at org.springframework.data.jpa.repository.query.AbstractStringBasedJpaQuery.doCreateQuery(AbstractStringBasedJpaQuery.java:145) ~[spring-data-jpa-4.1.0-M1.jar:4.1.0-M1]
```

oscarfanchin:
Hello @mp911de

Following your comment on the draft PR (https://github.com/spring-projects/spring-data-jpa/pull/4296#issuecomment-5117832861), I went on to explore a single-artifact version that runs on both runtimes. I had a feeling a Hibernate 7/8 adapter alone would not be enough, and it was not: I added a bridge for the EntityManager calls whose signature changed. For the rest it was enough to move to overloads that exist on both runtimes.

The implementation is in [#4307](https://github.com/spring-projects/spring-data-jpa/issues/4307). The same approach extends beyond Hibernate: the change is in the Jakarta Persistence API, not in the provider, so createNativeQuery(String, Class) will break the same way on EclipseLink once it ships Jakarta Persistence 4.

I routed the AOT code through the bridge as well. Generated repositories are compiled against the user's own classpath, so today they work either way. But the bridge is the one place where this behaviour can be corrected: if it ever has to carry a workaround, the generated repositories would silently miss it. It should not have to, but I would rather not depend on that.

I also added a reflective workaround for the tests that were failing before: Hibernate 8 no longer exposes untyped named queries through getNamedQueries. I had left those alone to keep reflection out, but the bridge already brings it in.

Let me know what you think. If you have a different direction in mind I am happy to explore it.

Issue #4187, opened 2026-02-11
Upgrade to Jakarta Persistence 4.0


mp911de:
Upgrade to the Jakarta Persistence API 4.0

- Replace `executeUpdate` calls on `Query`, to be replaced with `Statement.execute`
- Revisit for-removal deprecations on `Query` (`getResult*()` methods, `setCache*(…)` methods, `setLockMode()`, `setMaxResult`, `setFirstResult` methods)
- Investigate where we could use `TypedQuery` and introduce wrappers to defer calls to the `Query` interface

oscarfanchin:
Hi @mp911de Is there already a development branch for 4.0? I'd like to experiment.

PR #4296, opened 2026-07-06
Draft: Establish JPA 4 compatibility baseline with minimal changes

oscarfanchin:
## Summary

This draft provides an exploratory compatibility baseline for Jakarta Persistence 4 and Hibernate ORM 8.

The goal is to compile and run the existing Spring Data JPA functionality against this new baseline, validate the migration direction, and identify follow-up work.

This is not a complete or final support statement.

## Scope

- Minimal compatibility adaptations for Jakarta Persistence 4 and Hibernate ORM 8.
- Test adjustments required by API or provider behavior changes.
- Existing Spring Data JPA functionality only.

Support for new Jakarta Persistence 4 features is intentionally out of scope.

## Intentional exclusions

- EclipseLink tests are excluded because no Jakarta Persistence 4-compatible EclipseLink build is currently available.
- Querydsl tests are disabled because the Querydsl version currently used by the project is not compatible with Jakarta Persistence 4.

## Dependency baseline

This branch currently uses snapshot dependencies where required because the relevant Hibernate ORM and Jakarta Persistence APIs are still evolving.

More stable milestone builds are available, but they do not yet provide the compatibility surface required by this draft.

## Current status

- The project compiles.
- 3,591 tests were run: 3,413 passed, 108 were skipped, 7 failed, and 63 ended in errors.
- All 70 failures and errors are confined to three AOT repository test suites.
- All executed non-AOT tests pass, including the database-specific tests.
- The Spring Data Envers module test suite passes without failures.
- AOT named-query discovery requires dedicated follow-up work.

Test command:

mvn -pl spring-data-jpa -Pall-dbs clean process-test-classes \
  surefire:test@unit-test \
  surefire:test@integration-test \
  surefire:test@mysql-test \
  surefire:test@postgres-test \
  surefire:test@oracle-test \
  -Dmaven.test.failure.ignore=true -U

Full Surefire reports from the test run are attached.
[surefire-reports.zip](https://github.com/user-attachments/files/29703465/surefire-reports.zip)

## AOT note

The remaining AOT failures originate from named-query discovery in `QueriesFactory`.

With the current Hibernate ORM 8 baseline, untyped named queries may be represented by a `TypedQueryReference` whose result type is `null`.

`QueriesFactory` currently probes multiple result-type candidates, including `null`, when looking up named queries. Passing the `null` candidate to `EntityManagerFactory.getNamedQueries(Class<R>)` results in an NPE.

Removing the `null` candidate avoids the NPE but does not solve the underlying problem: untyped named queries can no longer be discovered. This causes AOT repository method contributions to be omitted and produces cascading metadata and integration-test failures.

This behavior needs further discussion and dedicated follow-up work, potentially involving coordination with Hibernate ORM.

## Related issues

Related to #4187  
Related to #4197

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/PartTreeJpaQuery.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/PartTreeJpaQuery.java
@@ -58,6 +58,7 @@ import org.springframework.util.Assert;
  * @author Jens Schauder
  * @author Mark Paluch
  * @author Сергей Цыпанов
+ * @author Oscar Fanchin
  */
 public class PartTreeJpaQuery extends AbstractJpaQuery {
 
@@ -125,10 +126,13 @@ public class PartTreeJpaQuery extends AbstractJpaQuery {
 		return queryPreparer.createQuery(accessor);
 	}
 
+	// SharedEntityManager may expose JPA 4 queries through a reduced proxy interface.
+	// Unwrap the underlying TypedQuery instead of casting the proxy directly.
+	// See https://github.com/spring-projects/spring-framework/issues/36878.
 	@Override
 	@SuppressWarnings("unchecked")
 	public TypedQuery<Long> doCreateCountQuery(JpaParametersParameterAccessor accessor) {
-		return (TypedQuery<Long>) countQuery.createQuery(accessor);
+		return (TypedQuery<Long>) (countQuery.createQuery(accessor).unwrap(TypedQuery.class));
 	}
 
 	@Override
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-jpa-parent</artifactId>
-	<version>4.2.0-SNAPSHOT</version>
+	<version>4.2.0-SNAPSHOT-JPA4</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data JPA Parent</name>
@@ -30,11 +30,12 @@
 		<antlr>4.13.2</antlr> <!-- align with Hibernate's parser -->
 		<eclipselink>5.0.0</eclipselink>
 		<eclipselink-next>5.0.0-SNAPSHOT</eclipselink-next>
-		<hibernate>7.4.0.Final</hibernate>
-		<hibernate-74-snapshots>7.4.1-SNAPSHOT</hibernate-74-snapshots>
+		<hibernate>8.0.0-SNAPSHOT</hibernate>
+		<hibernate-81-snapshots>8.1.0-SNAPSHOT</hibernate-81-snapshots>
 		<hsqldb>2.7.4</hsqldb>
 		<h2>2.3.232</h2>
-		<jakarta-persistence-api>3.2.0</jakarta-persistence-api>
+		<jakarta-persistence-api>4.0.0-M6</jakarta-persistence-api>
+		<jakarta-persistence-api-snapshot>4.0.0-SNAPSHOT</jakarta-persistence-api-snapshot>
 		<jsqlparser>5.3</jsqlparser>
 		<mysql-connector-java>9.7.0</mysql-connector-java>
 		<postgresql>42.7.11</postgresql>
@@ -44,6 +45,8 @@
 		<hibernate.groupId>org.hibernate</hibernate.groupId>
 		<antora-javadoc-artifactId>spring-data-jpa</antora-javadoc-artifactId>
 		<sonar.dynamicAnalysis>reuseReports</sonar.dynamicAnalysis>
+		<spring>7.1.0-SNAPSHOT</spring>
+		<jakarta-annotation-api>3.0.0</jakarta-annotation-api>
 	</properties>
 
 	<modules>
@@ -63,9 +66,10 @@
 			</repositories>
 		</profile>
 		<profile>
-			<id>hibernate-74-snapshots</id>
+			<id>hibernate-81-snapshots</id>
 			<properties>
-				<hibernate>${hibernate-74-snapshots}</hibernate>
+				<hibernate>${hibernate-81-snapshots}</hibernate>
+				<jakarta-persistence-api>${jakarta-persistence-api-snapshot}</jakarta-persistence-api>
 			</properties>
 			<repositories>
 				<repository>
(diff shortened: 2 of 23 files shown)


oscarfanchin:
Just a small follow-up: I intend to keep this PR in draft until Hibernate ORM 8 reaches the RC stage, since the underlying APIs are still evolving.

The goal of this PR is not to provide a complete and final Jakarta Persistence 4 migration at this stage. It is deliberately a minimal adaptation of the current baseline, intended to provide a working starting point for the subsequent changes and to avoid discovering major integration issues only at the last minute.

I’m keeping the branch aligned with the latest JPA 4.0 and Hibernate ORM 8 changes and periodically testing the integration as the APIs evolve.

In the meantime, some initial triage on the overall migration direction would be very useful. I’m not requesting a full code review yet; I would mainly like to understand whether this baseline approach is aligned with the Spring Data JPA roadmap and whether the AOT named-query discovery issue described above should be addressed here or coordinated with Hibernate ORM.

@mp911de  — since you opened #4187, some initial directional feedback on this baseline would be greatly appreciated.

Thank you for your attention and for all the work you do on the project.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/PartTreeJpaQuery.java
```diff
@@ -125,10 +126,13 @@ public Query doCreateQuery(JpaParametersParameterAccessor accessor) {
...
 	}
 
+	// SharedEntityManager may expose JPA 4 queries through a reduced proxy interface.
+	// Unwrap the underlying TypedQuery instead of casting the proxy directly.
+	// See https://github.com/spring-projects/spring-framework/issues/36878.
 	@Override
 	@SuppressWarnings("unchecked")
 	public TypedQuery<Long> doCreateCountQuery(JpaParametersParameterAccessor accessor) {
-		return (TypedQuery<Long>) countQuery.createQuery(accessor);
+		return (TypedQuery<Long>) (countQuery.createQuery(accessor).unwrap(TypedQuery.class));
```
mp911de, line 135: Nit: Adding a instanceof check would be helpful with a fallback to unwrap(…).

mp911de:
Thanks for having a look. The PR description is LLM-bloated and, especially the state on how many tests ran etc. 😉

A full move to a JPA 4.0 baseline requires a broader refactoring towards `TypedQuery`, as JPA's `Query` now carries several `@Deprecated(since = "4.0", forRemoval = true)` declarations.

I suggest keeping this change with a focus on #4197 to enable compatibility with Hibernate 7 and 8. Bridging the breaking changes between these versions requires some reflective glue. Consequently, our CI should run tests against both versions.

For now, the breaking changes appear to be limited to `HibernateUtil`, so we could introduce two reflective adapters for Hibernate 7 and 8, respectively.

oscarfanchin:
> Thanks for having a look. The PR description is LLM-bloated and, especially the state on how many tests ran etc. 😉
> 
> A full move to a JPA 4.0 baseline requires a broader refactoring towards `TypedQuery`, as JPA's `Query` now carries several `@Deprecated(since = "4.0", forRemoval = true)` declarations.
> 
> I suggest keeping this change with a focus on #4197 to enable compatibility with Hibernate 7 and 8. Bridging the breaking changes between these versions requires some reflective glue. Consequently, our CI should run tests against both versions.
> 
> For now, the breaking changes appear to be limited to `HibernateUtil`, so we could introduce two reflective adapters for Hibernate 7 and 8, respectively.

hi @mp911de 

First of all, thank you for the review and for taking the time to look into this.
I’d like to clarify which compatibility model you have in mind:

1. The same codebase is compiled separately against each baseline: one build against JPA 3.2, tested with Hibernate 7.4.x and EclipseLink 5.0.x, and another build against JPA 4.0, currently tested only with Hibernate 8.0. Each build is expected to work only with the baseline and providers it was compiled against, using minimal reflective adapters to isolate the changed APIs.

2. We continue compiling and releasing the artifact against the JPA 3.2 baseline, but the same binary is also expected to work at runtime with JPA 4.0 and Hibernate 8, using reflective adapters to bridge the differences.

Are you suggesting the first model, mainly to keep the codebase ready for the eventual baseline upgrade, or the second one, where the artifact compiled against JPA 3.2 must already support Hibernate 8 at runtime?
Before presenting this version of the PR, I already had something prepared for the first model. I have not yet explored the second scenario.

mp911de:
It is indeed the second variant. We do not intend to produce another artifact or version. Instead, the same Spring Data JPA version should support both Hibernate 7 with JPA 3.2 and Hibernate 8 with JPA 4.0.

For the November release, this approach leaves us with a JPA 3.2 baseline. Moving the baseline to JPA 4.0 and Hibernate 8.0, together with a compatible EclipseLink version, will require some more time until all involved parties are ready. There seems to be very little movement in EclipseLink who released their 5.0 GA version only just recently. I therefore  assume EclipseLink being one of the later JPA 4.0 adopters leaving us with the possibility to raise the JPA baseline only at a later time, maybe even with a Spring Data 5.0 release. I am happy to get more perspectives for better consideration.

oscarfanchin:
hi @mp911de, 
great, this is exactly what I expected would be needed and, honestly, what I was hoping for. It should be a very interesting challenge.
As soon as I have some free time, I will start from the other branch I already have locally. My plan is to keep the current profiles and add two more: one for the Hibernate 8.0 branch with JPA 4.0 M6, and one for Hibernate master, which has already moved to 8.1 and the latest JPA 4.0 snapshots. This should cover both Hibernate 8 lines while retaining the current Hibernate 7 setup. Does that sound like the right approach?
Given the change in scope and objective, I think it would be cleaner to open a separate PR.
The AOT issues I mentioned will probably remain, at least initially. Hibernate’s implementation of EntityManagerFactory.getNamedQueries(Class<?>) currently does not seem to return named queries without an explicitly declared result type when called with Object.class. Our current fallback then tries null, which results in an immediate NPE. This probably needs further investigation together with the Hibernate team.

spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/PartTreeJpaQuery.java
oscarfanchin: Taken on board, addressed in [#4307](https://github.com/oscarfanchin/spring-data-jpa/issues/4307).

oscarfanchin:
Closing this one: superseded by [#4307](https://github.com/spring-projects/spring-data-jpa/pull/4307), which takes the single-artifact direction. Discussion continues there.

Issue #3762, opened 2025-01-24
`FluentQuery.page(Pagable pagable)` does not consider `Pageable.sort`


Mrakobec:
it possible to implement the method `fluentQuery.page(Pageble pagable)` for automatically call `sortBy(Sort sort)` if `pagable.sort` is not null?

Mrakobec:
**Amazing! Thank you a lot!** 
Am I correct in seeing that I can specify only `.page()` for `<S extends T, R> R findBy(Specification<T> spec, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction)`, and the sorting will be applied? Additionally, if there was a `.sortBy()` specified, it will be overridden if sorting is present in the `pageRequest`?

Examples:

1. **Explicitly specifying `.sortBy()` before `.page()`:**
   ```java
   public Page<ExampleEntity> findAll(Specification<ExampleEntity> filters, PageRequest pageRequest) {
       return voucherRepository.findBy(filters, fetchable ->
               fetchable.project("someAttribute")
                       .sortBy(pageRequest.getSort()) // This will be overridden by pageRequest's sort
                       .page(pageRequest));
   }
   ```

2. **Only specifying `.page()` without `.sortBy()`:**
   ```java
   public Page<ExampleEntity> findAll(Specification<ExampleEntity> filters, PageRequest pageRequest) {
       return voucherRepository.findBy(filters, fetchable ->
               fetchable.project("someAttribute")
                       .page(pageRequest)); // Sorting from pageRequest will be applied
   }
   ```

mp911de:
Yes, that's right. We entirely override any previous `sortBy(…)` once a `Pageable` parameter is given.

PR #4298, opened 2026-07-06
Consider sort defined by sortBy(…) in fluent Specification slice and unpaged queries.

### LINKED ISSUE #4297: FetchableFluentQueryBySpecification ignores sortBy for slice and unpaged page
While building a Test API, I noticed that fluent `sortBy(…)` is ignored in some `JpaSpecificationExecutor.findBy(…)` paths when the provided `Pageable` itself is unsorted.

Reproducer: https://github.com/kmdy7991/jpa-issue-demo 

### Reproduction
```java
Slice<CustomerOrder> slice = orders.findBy(
		OrderSpecs.statusIs(Status.PAID),
		q -> q.sortBy(Sort.by(DESC, "createdAt"))
				.slice(PageRequest.of(0, 5)));
```

Expected: the query uses createdAt desc from sortBy(…).
Actual: the generated SQL has no order by:
```sql
select ... from orders co1_0
where co1_0.status=?
offset ? rows fetch first ? rows only
```

For comparison, the equivalent page(PageRequest.of(0, 5)) path applies the sort correctly:
```sql
select ... from orders co1_0
where co1_0.status=?
order by co1_0.created_at desc
offset ? rows fetch first ? rows only
```
The unpaged path shows the same issue:
```java
orders.findBy(
		OrderSpecs.amountAtLeast(20_000),
		q -> q.sortBy(Sort.by(DESC, "totalAmount"))
				.page(Pageable.unpaged()));
```

### Analysis

`FetchableFluentQueryBySpecification` uses `pageable.getSort()` in the affected paths, so an unsorted `Pageable` drops the sort configured through `sortBy(…)`.

The Querydsl counterpart uses `pageable.getSortOr(this.sort)`, which preserves the fluent sort when the `Pageable` is unsorted.

This looks related to GH-3762, which made `Pageable` sort override the fluent sort when present. The missing case here seems to be the fallback behavior: when the `Pageable` is unsorted, the fluent `sortBy(…)` sort should still be used. Existing tests cover the override case, but not this fallback for Specification `slice(…)` and unpaged terminals.

I can submit a pull request with the fix and regression tests.

kmdy7991:
`FetchableFluentQueryBySpecification` uses `pageable.getSort()` in `readSlice(…)` and in the unpaged branches of `slice(…)`, `page(…)`, and `page(…, countSpec)`, so a `Pageable` without a sort drops the sort configured through `sortBy(…)`. The Querydsl counterpart, `FetchableFluentQueryByPredicate`, uses `pageable.getSortOr(this.sort)` in the analogous page/slice paths and keeps the fluent sort as the fallback.

This change applies the same `getSortOr(…)` fallback to the Specification variant and adds regression tests for the unsorted-`PageRequest` slice path and the unpaged terminals. The existing GH-3762 tests only use `PageRequest`s that carry a sort (the override case), which is why the missing fallback went unnoticed.

Tests:
- `./mvnw.cmd -pl spring-data-jpa -Dtest=UserRepositoryTests test`

---

Closes #4297

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryBySpecification.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/FetchableFluentQueryBySpecification.java
@@ -54,6 +54,7 @@ import org.springframework.util.Assert;
  * @param <R> Result type
  * @author Greg Turnquist
  * @author Christoph Strobl
+ * @author kmdy7991
  * @since 3.0
  */
 class FetchableFluentQueryBySpecification<S, R> extends FluentQuerySupport<S, R>
@@ -175,18 +176,18 @@ class FetchableFluentQueryBySpecification<S, R> extends FluentQuerySupport<S, R>
 
 	@Override
 	public Slice<R> slice(Pageable pageable) {
-		return pageable.isUnpaged() ? new PageImpl<>(all(pageable.getSort())) : readSlice(pageable);
+		return pageable.isUnpaged() ? new PageImpl<>(all(pageable.getSortOr(this.sort))) : readSlice(pageable);
 	}
 
 	@Override
 	public Page<R> page(Pageable pageable) {
-		return pageable.isUnpaged() ? new PageImpl<>(all(pageable.getSort())) : readPage(pageable, spec);
+		return pageable.isUnpaged() ? new PageImpl<>(all(pageable.getSortOr(this.sort))) : readPage(pageable, spec);
 	}
 
 	@Override
 	@SuppressWarnings({ "rawtypes", "unchecked" })
 	public Page<R> page(Pageable pageable, Specification<?> countSpec) {
-		return pageable.isUnpaged() ? new PageImpl<>(all(pageable.getSort()))
+		return pageable.isUnpaged() ? new PageImpl<>(all(pageable.getSortOr(this.sort)))
 				: readPage(pageable, (Specification) countSpec);
 	}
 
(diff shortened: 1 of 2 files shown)


mp911de:
This is intentional behavior:

https://github.com/spring-projects/spring-data-jpa/blob/4a606f62caa578228119a8ee2373e29a87932c9b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/JpaSpecificationExecutor.java#L275-L281

and this is part of the contract:

https://github.com/spring-projects/spring-data-commons/blob/c64558c964c86903a3b9143b8961323d932058fe/src/main/java/org/springframework/data/repository/query/FluentQuery.java#L231-L235

With that in mind, we need to revise `FetchableFluentQueryByPredicate` to align with the contract. Pagination may introduce two potentially conflicting sort definitions. Since `Pageable` already encapsulates a `Sort`, it should override any previously configured sort.

kmdy7991:
@mp911de 

Thank you for the clarification — that makes sense, and I'm happy to rework this in that direction.

Two things I noticed while looking into what the alignment would touch, just to scope it correctly:

1. The JpaSpecificationExecutor Javadoc for page(Pageable, PredicateSpecification) currently states the conditional rule ("…will override any previously specified sort if the Sort object is not Sort#isUnsorted()"), so it would need updating to the unconditional wording along with the code.
2. readPage(…) in FetchableFluentQueryBySpecification also uses pageable.getSortOr(this.sort) today, and a few existing tests assert the fallback behavior (findByFluentSpecificationPageCustomCountSpec / GH-3727 with an unsorted PageRequest, and findByFluentPredicateSlice / GH-3764 on the Querydsl side) — those would need adjusting as well.

Would you like me to rework this PR to do that — align FetchableFluentQueryByPredicate (and readPage(…)) on pageable.getSort(), update the Javadoc, and adjust the affected tests — or would you prefer to handle it on your side? Either way works for me.

PR #4335, opened 2026-08-27
Allow CAST as a state field name in EQL

wantaekchoi:
`Eql.g4` declares a `CAST` token for the cast function rules but leaves it out of `reserved_word`, so an entity attribute named `cast` cannot be parsed. `select m.cast from Movie m where m.cast is not null` fails with a `BadJpqlGrammarException` (`At 1:9 and token 'cast', no viable alternative`), while JPQL and HQL parse the same query. Applications running on EclipseLink are affected because `DefaultQueryEnhancerSelector` picks the EQL parser when Hibernate is absent.

`Jpql.g4` gained the `reserved_word` entry in b0a29229c along with the reworked cast function rules. `Eql.g4` received the rules but not the `reserved_word` entry, and it has never had one. `CAST` is the only token that one grammar lists there and the other does not; the remaining entries are identical in both.

This PR adds `CAST` to the EQL `reserved_word` rule and covers the shape in the shared renderer TCK, so JPQL, EQL, and HQL all assert it.

Verified with `./mvnw -pl spring-data-jpa -Dtest='*QueryRenderer*,*Eql*,*Jpql*,*Hql*' test`. Removing the grammar line again leaves the new test as the only failure.

### DIFF
--- a/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
+++ b/spring-data-jpa/src/main/antlr4/org/springframework/data/jpa/repository/query/Eql.g4
@@ -24,6 +24,7 @@ grammar Eql;
  * @author Greg Turnquist
  * @author Christoph Strobl
  * @author Jewoo Shin
+ * @author Wantaek Choi
  * @since 3.2
  */
 }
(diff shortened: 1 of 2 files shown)


mp911de:

PR #4343, opened 2026-09-03
Fix count query derivation for DISTINCT selections containing expressions

### LINKED ISSUE #4341: Derived count query for `SELECT DISTINCT` mangles select expressions into invalid JPQL
Deriving a count query from a `SELECT DISTINCT` query whose selection contains expressions produces invalid JPQL in `QueryTransformers.CountSelectionTokenStream`:

1. Flattening the selection tokens drops all whitespace of multi-token expressions.
2. `AS` (plus the following token) is stripped at any parenthesis nesting level, breaking `cast(x as string)`.
3. `withoutConstructorExpression()` removes **all** parentheses rather than just the constructor wrapper, breaking function calls inside constructor arguments.

Examples on `main` (also affects JPQL and EQL; HQL is affected for the constructor case):

```text
select distinct coalesce(u.name, u.lastname) from User u where u.foo = ?1
→ count query: select count(distinct coalescename, lastname) from User where foo = ?1

select distinct case when u.age > 18 then 'adult' else 'minor' end from User u
→ count query: select count(distinct casewhenu.age>18then'adult'else'minor'end) from User u
```

Both derived count queries are rejected by the project's own query parsers (`BadJpqlGrammarException: no viable alternative`), so any paged repository query of this shape fails at count-query execution.

Expected: the derived count query preserves the expression's spacing, keeps `AS` inside function calls, and retains nested parentheses when unwrapping a constructor expression. I will submit a PR with a fix and tests for all three dialects.

harrisleesh:
Closes #4341

Deriving a count query from a `SELECT DISTINCT` query whose selection contains expressions produced invalid JPQL: flattening the selection dropped all whitespace of multi-token expressions (`count(distinct casewhenu.age>18then...)`), `AS` was stripped at any parenthesis nesting level (breaking `cast(x as string)`), and `withoutConstructorExpression()` removed all parentheses instead of only the constructor wrapper (`count(distinct coalescename, lastname)`). The derived queries are rejected by the project's own parsers, so paged repository queries of this shape fail at count-query execution.

This change strips `AS` only at nesting level zero, retains nested parentheses when unwrapping a constructor expression, and reconstructs the spacing of the flattened selection tokens.

Five regression tests added across the HQL/JPQL/EQL transformer tests (all fail on `main`, pass with the fix); the full transformer test suites pass (HQL 102, JPQL 82, EQL 82) and the module unit test suite is green.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryTransformers.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryTransformers.java
@@ -44,6 +44,7 @@ class QueryTransformers {
 			List<QueryToken> target = new ArrayList<>(selection.size());
 			boolean skipNext = false;
 			boolean containsNew = false;
+			int nestingLevel = 0;
 
 			for (QueryToken token : selection) {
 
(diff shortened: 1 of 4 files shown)


mp911de:
This isn't a proper approach. We've got a distinction between expressions (that require isolation by white space) and regular tokens that can be concatenated without extra whitespaces. The change is introducing another set of rules and that makes the overall composition more complex with leaving the actual problem unsolved.

PR #4344, opened 2026-09-03
Apply Sort only to the outermost query for parenthesized HQL set operations

### LINKED ISSUE #4342: Sort is applied to every arm of a parenthesized HQL set operation
When a sorted query is applied to an HQL query using a parenthesized set operation, `HqlSortedQueryTransformer.visitNestedQueryExpression` re-enters `visitOrderedQuery(ctx)`, which applies the transformer's `Sort` again at every nesting level — although `visitQueryExpression` explicitly passes `Sort.unsorted()` to non-last arms.

On `main`, sorting

```sql
(SELECT tb FROM Test tb WHERE tb.type = 'A') UNION (SELECT tb FROM Test tb WHERE tb.type = 'B')
```

by `tb.Type` produces:

```sql
(SELECT tb FROM Test tb WHERE tb.type = 'A' order by tb.Type asc)UNION (SELECT tb FROM Test tb WHERE tb.type = 'B' order by tb.Type asc) order by tb.Type asc
```

The `ORDER BY` is applied three times. When the arms use different aliases — e.g. `(select u from User u) union (select v from Visitor v)` with `Sort.by("name")` — the `order by u.name` injected into the second arm references an unknown alias and Hibernate rejects the query; even with equal aliases the per-arm ordering changes set-operation semantics.

Expected: `Sort` is applied exactly once, at the outermost query level, matching the invariant already asserted for FROM-subqueries. I will submit a PR with a fix and tests.

harrisleesh:
Closes #4342

Rendering a parenthesized query expression re-entered `visitOrderedQuery`, which applied the `Sort` again at every nesting level, so the `ORDER BY` was injected into each arm of a set operation such as `(SELECT ...) UNION (SELECT ...)` in addition to the outer query — three times in total for a two-arm union. With different arm aliases the injected ordering references an unknown alias and Hibernate rejects the query; with equal aliases it changes set-operation semantics.

`HqlSortedQueryTransformer` now tracks whether the visitor is inside a nested query expression and substitutes `Sort.unsorted()` while nested, so the `Sort` is applied exactly once at the outermost level — consistent with the invariant already asserted for FROM-subqueries (GH-3427 test).

Two regression tests added (both fail on `main`, pass with the fix); `HqlQueryTransformerTests` passes 101/101.

### DIFF
--- a/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlSortedQueryTransformer.java
+++ b/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/HqlSortedQueryTransformer.java
@@ -43,6 +43,13 @@ class HqlSortedQueryTransformer extends HqlQueryRenderer {
 	private final @Nullable String primaryFromAlias;
 	private final @Nullable DtoProjectionTransformerDelegate dtoDelegate;
 
+	/**
+	 * Whether the visitor is currently within a parenthesized (nested) query expression. {@link Sort} must only be
+	 * applied to the outermost query level, not to individual arms of a set operation such as
+	 * {@code (SELECT …) UNION (SELECT …)}.
+	 */
+	private boolean nestedQueryExpression = false;
+
 	HqlSortedQueryTransformer(Sort sort, HibernateQueryInformation queryInformation,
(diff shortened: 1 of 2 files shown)


mp911de:

PR #4360, opened 2026-09-19
Allow modulo operator `%` in HQL expressions.

### LINKED ISSUE #4359: HQL does not support the modulo operator (%)
The HQL parser rejects the `%` (modulo) operator, although Hibernate accepts it.

```java
@Query("select c from Call c where c.duration % 2 = 0")
List<Call> findEvenDurations();
```

fails with:

```
org.springframework.data.jpa.repository.query.BadJpqlGrammarException:
At 1:38, token recognition error at: '%';
Bad HQL grammar [select c from Call c where c.duration % 2 = 0]
```

`Hql.g4` only allows `*` and `/` as multiplicative operators, while Hibernate's grammar also includes `%`:

```antlr
// Hql.g4
expression
    | expression op=('*' | '/') expression    # MultiplicationExpression
```

```antlr
// Hibernate HqlParser.g4
multiplicativeOperator
	: SLASH
	| PERCENT_OP
	| ASTERISK
	;
```

Running the same queries through Hibernate's own `HqlParser` and through `JpaQueryEnhancer.HqlQueryParser`:

| query | Hibernate 7.4 | Spring Data JPA |
|---|---|---|
| `select b.a % b.b from B b` | parses | **rejected** |
| `select b from B b where b.n % 2 = 0` | parses | **rejected** |
| `select mod(b.a, b.b) from B b` | parses | parses |

The same applies to `JpaSort.unsafe("age % 2")`, which is parsed with the HQL grammar as well. Note that `HqlOrderExpressionVisitor.visitMultiplicationExpression` currently maps every operator other than `*` to `CriteriaBuilder.quot(…)`, 
so it needs to map `%` to `CriteriaBuilder.mod(…)` once the grammar accepts it.

ohchanKyu:
Hibernate's HQL grammar accepts `%` as a multiplicative operator, whereas Spring Data JPA's grammar only
allowed `*` and `/`, so queries such as `c.duration % 2 = 0` were rejected with a `BadJpqlGrammarException`. 
Add `%` to the multiplication expression and map it to `CriteriaBuilder.mod(…)` in `HqlOrderExpressionVisitor`,
 which previously treated every operator other than `*` as a division.

Closes #4359


christophstrobl:
Thank you @ohchanKyu - LGTM 👍

